PR #2, opened 2016-08-29
Adds Closed Sets

mronian:


mronian:
cc @timholy , @ararslan :)

src/closed.jl
```diff
@@ -0,0 +1,45 @@
...
+	left, right = promote(left, right)
+	ClosedSet(left, right)
+end
+
+function union(A::ClosedSet, B::ClosedSet)
+	left = min(A.left, B.left)
+	right = max(A.right, B.right)
+	left, right = promote(left, right)
+	ClosedSet(left, right)
+end
```
mronian: Should `union` and `intersect` if any of the sets is empty?

src/closed.jl
```diff
@@ -0,0 +1,44 @@
...
+	left::T
+	right::T
+end
+
+function ClosedSet(left, right)
+	(I_left, I_right) = promote(left, right)
+	ClosedSet(I_left, I_right)
+end
+
+show(io::IO, I::ClosedSet) = print(io, "[", I.left, ",", I.right, "]")
```
mronian: How to test IO?

src/closed.jl
```diff
@@ -0,0 +1,44 @@
+"""
+Defines the ClosedSet type.
+
+A ClosedSet(left, right) is an interval set where both bounds are closed i.e. the range is [left, right]
+"""
+immutable ClosedSet{T}
+	left::T
```
timholy: indentation 4 chars

src/closed.jl
```diff
@@ -0,0 +1,44 @@
...
+end
+
+function ClosedSet(left, right)
+	(I_left, I_right) = promote(left, right)
+	ClosedSet(I_left, I_right)
+end
+
+show(io::IO, I::ClosedSet) = print(io, "[", I.left, ",", I.right, "]")
+
+length(I::ClosedSet) = I.right - I.left + 1
```
timholy: Doesn't work for non-integers. Maybe we shouldn't define `length` at all.

src/closed.jl
```diff
@@ -0,0 +1,44 @@
...
+
+in(v, I::ClosedSet) = I.left <= v <= I.right
+
+isempty(A::ClosedSet) = A.left > A.right
+
+isequal(A::ClosedSet, B::ClosedSet) = (A.left == B.left && A.right == B.right) || isempty(A) && isempty(B)
```
timholy: Would be worth benchmarking whether this version or the non-shortcircuiting version (`&` instead of `&&` and `|` instead of `||`) is faster.

src/closed.jl
```diff
@@ -0,0 +1,44 @@
...
+
+==(A::ClosedSet, B::ClosedSet) = isequal(A, B)
```
timholy: Do you need to define this? Why is `==` defined in terms of `isequal`?

src/closed.jl
```diff
@@ -0,0 +1,44 @@
...
+
+function intersect(A::ClosedSet, B::ClosedSet)
+	left = max(A.left, B.left)
+	right = min(A.right, B.right)
+	left, right = promote(left, right)
```
timholy: You don't need this line, because the constructor will take care of that. Moreover, `left` and `right` should already be of the same type as each other, even if they differ from the type of `A` and/or `B`, because `min` and `max` promote.

test/runtests.jl
```diff
@@ -1,5 +1,39 @@
-using IntervalSets
-using Base.Test
+module IntervalSetsTest
 
-# write your own tests here
-@test 1 == 2
+using IntervalSets, Base.Test, FactCheck
```
timholy: Would it be OK to switch to Base.Test and `@testset`?

src/closed.jl
timholy: `intersect` doesn't have to, just add some tests for intersection of empty sets and I bet you'll see it already works as-is.

More importantly, `union` needs to make sure that the intervals intersect; I think this implementation will "fill in" gaps between two disjoint intervals. It should throw an error (`InexactError`? `ArgumentError`?) when the two inputs don't intersect, because the union can't be expressed as a single closed set.

src/closed.jl
timholy: See `IOBuffer` (and many examples in Julia's own `test/` folder).

REQUIRE
```diff
@@ -1 +1 @@
-julia 0.4
+julia 0.5
```
timholy: At present I think you need 0.5-, since 0.5 isn't released.

.travis.yml
```diff
@@ -1,19 +1,16 @@
...
 language: julia
 os:
-  - linux
-  - osx
+    - linux
+    - osx
 julia:
-  - release
-  - nightly
+    - nightly
```
timholy: Better make this
```
- 0.5
- nightly
```
because those two are already different from one another (and the differences will keep growing).

.travis.yml
```diff
@@ -1,19 +1,16 @@
...
 notifications:
-  email: false
-# uncomment the following lines to override the default test script
-#script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("IntervalSets"); Pkg.test("IntervalSets"; coverage=true)'
+    email: false
+script:
+    - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+    - julia -e 'Pkg.clone(pwd()); Pkg.build("IntervalSets")'
```
timholy: Any reason to uncomment the original? I don't think this does anything different.

src/closed.jl
```diff
@@ -0,0 +1,44 @@
...
+end
+
+function union(A::ClosedSet, B::ClosedSet)
+	left = min(A.left, B.left)
+	right = max(A.right, B.right)
+	left, right = promote(left, right)
+	ClosedSet(left, right)
+end
+
+issubset(A::ClosedSet, B::ClosedSet) =  isempty(A) || (A.left in B && A.right in B)
```
timholy: Likewise here about the short-circuiting benchmarking.

src/closed.jl
ararslan: More specifically, 4 spaces. Indenting with 4 tabs would be chaos. 😄

.travis.yml
```diff
@@ -1,19 +1,16 @@
-# Documentation: http://docs.travis-ci.com/user/languages/julia/
 language: julia
 os:
-  - linux
-  - osx
+    - linux
```
ararslan: It doesn't really matter but out of curiosity why change the indentation style in the default Travis file?

mronian: I copied the entire contents from a travis file in another package which probably lead to this :sweat_smile: I'll change it :smile:

.travis.yml
mronian: I copied the entire contents from a travis file in another package :sweat_smile:, which lead to these changes. Sorry about that! Will restore them :smile:

.travis.yml
timholy: Just checking, you know about `PkgDev.generate`, right?

mronian: Yep. I'll clean this mess :smiley:

src/closed.jl
mronian: Should `==` be only `(A.left == B.left && A.right == B.right)`  and not the `isempty` part of `isequal`?

src/closed.jl
mronian: This becomes faster if `|` and `&` are used. (First benchmark is for the original one with `||` and `&&`)- 

```julia
julia> @benchmark issubset(A, B)
julia> @benchmark issubset(A, B)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     995
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     27.00 ns (0.00% GC)
  median time:      27.00 ns (0.00% GC)
  mean time:        27.70 ns (0.00% GC)
  maximum time:     71.00 ns (0.00% GC)

BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     997
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     22.00 ns (0.00% GC)
  median time:      22.00 ns (0.00% GC)
  mean time:        22.40 ns (0.00% GC)
  maximum time:     49.00 ns (0.00% GC)
```

src/closed.jl
mronian: This becomes slower with the non-shortcircuiting version. (First benchmark is for the one with `||` and `&&`) - 

```julia
julia> @benchmark isequal(A, B)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     998
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     16.00 ns (0.00% GC)
  median time:      16.00 ns (0.00% GC)
  mean time:        16.49 ns (0.00% GC)
  maximum time:     43.00 ns (0.00% GC)
julia> @benchmark isequal(A, B)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     997
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     21.00 ns (0.00% GC)
  median time:      21.00 ns (0.00% GC)
  mean time:        21.91 ns (0.00% GC)
  maximum time:     66.00 ns (0.00% GC)
```

src/closed.jl
timholy: Hmmm, interesting. Not entirely sure, but I wonder if `0.0 == -0.0` is justification for saying that all empty sets should return `true` for `==`.

It seems kind of odd that `==` is more permissive than `isequal` about the sign of 0 but less permissive about `NaN`.

src/closed.jl
timholy: Better to use `$`, i.e., `@benchmark issubset($A, $B)`.

src/closed.jl
timholy: Oh, and this should probably print as `left..right` since that's how they will be constructed (that way you can copy/paste output into the REPL and have it work).

src/closed.jl
timholy: Case of floats is interesting because you should check against `nextfloat` (when the edges are touching but not overlapping).

src/closed.jl
mronian: Shall I remove the `==` function?

src/closed.jl
```diff
@@ -0,0 +1,54 @@
+"""
+Defines the ClosedSet type.
+
+A ClosedSet(left, right) is an interval set where both bounds are closed i.e. the range is [left, right]
```
ararslan: "Defines the ClosedSet type" is probably unnecessary since it should (hopefully) be obvious to a user that the type constructor constructs the type. :smile: 

I'd recommend putting backticks around `ClosedSet(left, right)` and perhaps rephrasing a bit as: "A `ClosedSet(left, right)` is an interval set that includes both its upper and lower bounds. In mathematical notation, the constructed range is `[left, right]`."

I say "in mathematical notation" there so as to avoid confusion with the syntax for arrays.

src/closed.jl
```diff
@@ -0,0 +1,54 @@
...
+==(A::ClosedSet, B::ClosedSet) = isequal(A, B)
+
+function intersect(A::ClosedSet, B::ClosedSet)
+    left = max(A.left, B.left)
+    right = min(A.right, B.right)
+    ClosedSet(left, right)
+end
+
+function union{T<:AbstractFloat}(A::ClosedSet{T}, B::ClosedSet{T})
+    @assert max(A.left, B.left) <= nextfloat(min(A.right, B.right)) throw(ArgumentError("The sets are disjoint."))
```
ararslan: Rather than using `@assert`, I'd recommend the `condition || throw(...)` syntax. I'd also maybe make the error more specific, maybe something like `Cannot construct the union of disjoint sets`?

src/closed.jl
timholy: No, on balance I think we should test equality, it just should be
```
==(A::ClosedSet, B::ClosedSet) = (A.left == B.left & A.right == B.right) | (isempty(A) & isempty(B))
```
or possibly the short-circuiting version. (I doubt it, though.)

src/closed.jl
```diff
@@ -0,0 +1,54 @@
...
+Defines the ClosedSet type.
+
+A ClosedSet(left, right) is an interval set where both bounds are closed i.e. the range is [left, right]
+"""
+immutable ClosedSet{T}
+    left::T
+    right::T
+end
+
+function ClosedSet(left, right)
```
timholy: This can be a one-liner, if you wish. `@code_llvm` suggests that the 2-argument case does not suffer from splatting penalty (perhaps it's special-cased in the compiler).

src/closed.jl
timholy: Oh, I see now, you have both `isequal` and `==`. It's hard to say which of these we should define, but I lean towards `==`. Or maybe both with `isequal` using `isequal` for comparing `left` and `right`, and `==` using `==` for the same comparisons.

timholy:
Looking quite good (obviously we'll need some more documentation at some point). Nice work!

src/closed.jl
simonbyrne: Also, why `ClosedSet` and not `ClosedInterval`?

ararslan: Simon makes a very good point. Formally speaking, a "closed set" is an *arbitrary* set that contains its limit points whereas a closed interval is much more specific (and is what is implemented).

timholy: I guess there's a slight conflict with https://github.com/scheinerman/ClosedIntervals.jl/blob/master/src/ClosedIntervals.jl, though. Still, it's undeniably the better term, so I guess we should use module names to disambiguate.

ararslan: I'm not super worried about the conflict with ClosedIntervals.jl since it isn't METADATA-registered (and would take some work to get ready for registration) nor does it appear to be all that actively developed.

mronian: I had the same concern as @timholy mentioned. But I guess the consensus is to use `ClosedIntervals` .

mronian:
Did some benchmarks and it turns out that for `==`, the version with `&, |` is faster. For other functions it is the other way round. I have kept whichever is faster.

src/closed.jl
```diff
@@ -0,0 +1,51 @@
...
+
+show(io::IO, I::ClosedInterval) = print(io, I.left, "..", I.right)
+
+in(v, I::ClosedInterval) = I.left <= v <= I.right
+
+isempty(A::ClosedInterval) = A.left > A.right
+
+isequal(A::ClosedInterval, B::ClosedInterval) = (isequal(A.left, B.left) & isequal(A.right, B.right)) | isempty(A) & isempty(B)
+
+==(A::ClosedInterval, B::ClosedInterval) = (A.left == B.left && A.right == B.right) || isempty(A) && isempty(B)
```
mronian: Is this fine?

mronian:
What is an appropriate time to register a package? (I want to use this in other packages, but considering it has only `ClosedInterval` right now, will it be accepted?)

ararslan:
In terms of registration, I think as long as some basic set of functionality is implemented (and that's reflected in the version number, e.g. 0.0.1) then whenever things are working is a fine time.

src/closed.jl
timholy: Just checking, does the operator precedence work out? Might want to add a test `I == I` and/or `L == L` below.

timholy:
Agreed, I don't think there's a minimum unit of functionality required for a package.

PR #3, opened 2016-09-11
Add features for supporting AxisArrays

timholy:
In preparation for rebasing [AxisArrays](https://github.com/mbauman/AxisArrays.jl) on this interval type, I needed to add a number of new features.

Perhaps the only "controversial" feature is the ability to construct a `ClosedInterval` from a single number:

```julia
julia> ClosedInterval(7)
7..7
```

This was supported in AxisArrays; I'm not certain it's a good idea, but I thought I'd throw it in to start the discussion. Locally I've already used this successfully in AxisArrays, and I note that if I comment out [this line](https://github.com/mbauman/AxisArrays.jl/blob/08f266dce5a68844e7ee7d7ab630ffc820ba7f07/test/intervals.jl#L87), all of `AxisArrays` tests pass even without this feature. So I'm not convinced it's important, but perhaps @mbauman can shed some light on something I may be missing.

CC @mronian.

### DIFF
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -13,4 +13,8 @@ include("closed.jl")
 ordered{T}(a::T, b::T) = ifelse(a < b, (a, b), (b, a))
 ordered(a, b) = ordered(promote(a, b)...)
 
+checked_conversion{T}(::Type{T}, a, b) = _checked_conversion(T, convert(T, a), convert(T, b))
+_checked_conversion{T}(::Type{T}, a::T, b::T) = a, b
+_checked_conversion{T}(::Type{T}, a, b) = throw(ArgumentError("$a and $b are not both of type $T"))
+
 end # module
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -4,11 +4,23 @@ module IntervalSets
 
 # package code goes here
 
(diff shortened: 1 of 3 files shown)


mronian:
This isn't that intuitive though. Is there any significance of creating an interval with a single number?

One more thing I forgot to discuss earlier was the `..` notation. Once OpenInterval, LOpen and Ropen are implemented, what notation do we use for those?

timholy:
From the standpoint of sets, I don't think it's important and agree it's potentially confusing. If you're thinking about an interval in terms of numerics, perhaps it makes more sense. But AFAICT AxisArrays doesn't actually use that feature, so even there is may not be necessary. In which case I'd have some preference for stripping it out.

With regard to other interval types, I'm not sure there's enough compact notation for them all. It seems that closed intervals are the most useful type, so it's nice to have compact syntax there at least. That said, I'm all in favor of nice notation for the others, too, if we can find something sensible.

timholy:
I decided to delete the single-argument constructors. They can be added later, of course.

mbauman:
This seems very sensible.

My memory now is very fuzzy, but I think there may have been a point where I relied upon conversion from a scalar to a closed integer.  That case (if it's needed at all, which I'm not sure about) is much better handled by conversion rather than construction.

PR #4, opened 2016-09-11
Add content to the README

timholy:


README.md
```diff
@@ -4,3 +4,62 @@ Interval Sets for Julia
...
+0.5..2.5
+```
+
+The `±` operator may be typed as `\pm<TAB>` (using Julia's LaTeX
+syntax tab-completion).
+
+Intervals also support the expected set operations:
+
+```julia
+julia> 1.75 ∈ 1.5±1  # can also use `in`
```
mronian: Should we also mention how to get the `∈` and `∪` characters?

timholy: fixed

PR #6, opened 2016-11-17
Added width (with tests).

tpapp:
Trivial addition for the width of the interval. Implementation also
handles case when type of the width is different from the eltype (eg
ClosedInterval{Date}).


ararslan:
Looks good, thanks! It might be nice to also include a test or two that cover the case of homogeneous types, e.g. `Int`.

tpapp:
@ararslan: I added extra tests, but `PkgDev.submit` created another PR (#7) instead of updating this one, sorry for the mess.

ararslan:
No problem. I'll go ahead and close this one since the other PR includes the changes from this one.

PR #7, opened 2016-11-18
added width (updated)

tpapp:
An update of the previous pull request, with extra tests as suggested by @ararslan.


mronian:
Thanks @tpapp :)

PR #10, opened 2017-02-18
Fix typealias and inner constructor depwarns on 0.6

timholy:


REQUIRE
```diff
@@ -1 +1,2 @@
-julia 0.5-
+julia 0.5
+Compat 0.18
```
tkelman: 0.17

PR #12, opened 2017-03-12
Support conversions from Ranges

timholy:


test/runtests.jl
```diff
@@ -81,3 +82,5 @@ using Base.Test
         @test width(ClosedInterval(4.0,8.0)) ≡ 4.0
     end
 end
+
+nothing
```
ararslan: What's this for?

timholy: Just prevents output if you run the tests at the REPL prompt.

Issue #21, opened 2017-10-19
width of integer & similar intervals


tpapp:
Currently
```julia
julia> width(Date(1980,1,1)..Date(1980,1,10))
9 days

julia> width(1..10)
9
```
which makes sense in some contexts, but in others `10` would be a better choice (eg duration of a spell represented by the interval).

I would like to initiate discussion on how to handle this. The following come to mind:

1. Encourage the user to redefine `width` with `+1` or equivalent whenever that makes sense, but don't do this automatically. *Problem*: `width(1.0..10.0) ≠ width(1..10)` if one redefines it. For `Date` and similar this would not be a problem.

2. To implement 1. more elegantly, one can define a trait, eg `Date` would have the trait `RepresentsDuration()`, `width` would test for that, and add `1` accordingly. The default would be not having this trait. *Problem*: same as above, *advantage*: user could inspect this behavior for generic types.

3. Add another generic function, eg `span`, which adds `1` whenever applicable. Again, with or without trait.

Thoughts?

timholy:
How about `length` and define it only for `Integer` intervals?

tpapp:
Thanks for the suggestion, see PR https://github.com/JuliaMath/IntervalSets.jl/pull/22.

dlfivefifty:
We now have `duration`. But probably the current behaviour is wrong and duration of dates should return a date?
```julia
julia> duration(Date(1980,1,1)..Date(1980,1,10))
10

julia> duration(1..10)
10
```

timholy:
Yes, definitely the units need to be preserved. And just to check (sorry, short on time), I think we got rid of `length`, right?

dlfivefifty:
Yes we got rid of `length`: `length` is a notion of cardinality, not measure, so it was an abuse of terminology, and also doesn't play well with Base's assumptions on length.

If anything, the following is the most reasonable definition of `length:
```julia
length(d::AbstractInterval) = (@assert !isempty(d); ℵ₁)`
```

hyrodium:
`length` was removed https://github.com/JuliaMath/IntervalSets.jl/issues/31, so I'll close this issue. Note that `duration` will also be removed in the future release. (#102)

PR #22, opened 2017-10-19
Added methods for Base.length.

tpapp:
See discussion at https://github.com/JuliaMath/IntervalSets.jl/issues/21, this should fix the issue.

PR aims to maintain support for v0.5, too, hence the `@compat` for `Union` and the old syntax for type parameters (hope I did it right).


tpapp:
Closing in favor of another PR -- length should return integers.

PR #26, opened 2018-03-19
Add open and mixed open/closed intervals

dlfivefifty:
This PR adds support for general closed/open endpoints.

It also makes the following changes which I believe are necessary:

1. `union(a .. b, nextfloat(b) .. c)` now throws an error (for `a < b < c`): this is consistent with the fact that `union(0 .. 1, 2..3)` should not be `0 .. 3`, and because 
```julia
a, b, c = 0.0, 1.0, 2.0
x = nextfloat(BigFloat(b))
!(x in a..b) && !(x in nextfloat(b)..c)
```
Previously `x in union(a .. b, nextfloat(b) .. c)`  returned true, which is inconsistent with the above!
2. `convert(UnitRange, 0..1)` is now deprecated in favour of `UnitRange(0..1)`, as `convert` should throw an error (e.g. `convert(Int, 1+Im)` throws an error).


dlfivefifty:
@timholy This is ready for review.

src/IntervalSets.jl
```diff
@@ -4,29 +4,73 @@ module IntervalSets
 
 using Base: @pure
 import Base: eltype, convert, show, in, length, isempty, isequal, issubset, ==, hash,
-             union, intersect, minimum, maximum, extrema, range, ⊇
+             union, intersect, minimum, maximum, extrema, range, ⊇, mean, median
```
ararslan: In Julia 0.7, `mean` and `median` are in the Statistics stdlib package. So instead of adding them here, you could add
```julia
import Compat.Statistics: mean, median
```
Just make sure that REQUIRE has an appropriate lower bound for Compat in that case.

src/IntervalSets.jl
```diff
@@ -4,29 +4,73 @@ module IntervalSets
...
+
+function supremum(d::AbstractInterval{T}) where T
+    a = leftendpoint(d)
+    b = rightendpoint(d)
+    a > b && throw(ArgumentError("Supremum not defined for empty intervals"))
+    b
+end
+
+mean(d::AbstractInterval) = one(eltype(d))/2 * (leftendpoint(d) + rightendpoint(d))
+median(d::AbstractInterval) = mean(d)
```
ararslan: I'm not sure about this definition. Typically (not always, granted) the median of some set is an element of the set, but that isn't guaranteed when using a mean. Maybe this is fine, I dunno.

src/IntervalSets.jl
```diff
@@ -4,29 +4,73 @@ module IntervalSets
...
+end
+
+function supremum(d::AbstractInterval{T}) where T
+    a = leftendpoint(d)
+    b = rightendpoint(d)
+    a > b && throw(ArgumentError("Supremum not defined for empty intervals"))
+    b
+end
+
+mean(d::AbstractInterval) = one(eltype(d))/2 * (leftendpoint(d) + rightendpoint(d))
```
ararslan: Is the `one` part needed here? Couldn't you just write this as
```julia
(leftendpoint(d) + rightendpoint(d)) / 2
```
?

dlfivefifty:
@ararslan I've addressed your comments, and just removed `median` as I don't see any use for it. I've also made the following changes which you might have some opinions on:

1. I renamed `AbstractInfiniteSet` as `Domain`, as `AbstractSet` has a (currently undefined) meaning in Julia, which may implicitly imply that the set is countable. I'd be fine removing this abstract type completely.
2. Added abstract type `TypedEndpointsInterval{L,R,T}` to allow custom implementations of intervals such as `MyClosedUnitInterval`.
3. Deprecated `length` in favour of `duration`, to address  #31. This PR already implemented `width` for `arclength`.

ararslan:
Looks good to me, but I'm not a primary maintainer of this package, so it'd be good to hear from someone else as well before merging.

dlfivefifty:
We can always chat next week about it...

dlfivefifty:
Is `leftendpoint`/`rightendpoint` too verbose? Would `left` and `right` be better?

ararslan:
Personally I prefer the clarity of `(left|right)endpoint`; just `left` and `right` seem overly general, IMO.

dlfivefifty:
@timholy I mentioned this PR at JuliaCon. Do you have time to review it now? 

(Still no rush on my part but a revamp of domains in ApproxFun.jl is waiting on this.)

timholy:
In other words, merge at will. If you are (justifiably) annoyed by the delay, I can help fix those merge conflicts. Just let me know.

dlfivefifty:
That's OK, I've resolved the conflicts, though you might have comments on the changes I made:

1. I removed the general `AbstractUnitRange` conversion because it seemed to broad, in favour of specific ones:
```julia
UnitRange{T}(i::TypedEndpointsInterval{:closed,:closed,I}) where {T<:Integer,I<:Integer} = UnitRange{T}(minimum(i), maximum(i))
UnitRange(i::TypedEndpointsInterval{:closed,:closed,I}) where {I<:Integer} = UnitRange{I}(i)
```
2. I added `convert(AbstractInterval, x::Number)` and variants
3. I overrode `issubset`, as the existing definition called the now deprecated `length`.

timholy:
On my phone. See my recent PR to Base re issubset. Can't really check rest
now, you are authorized to decide😀

timholy:
Having looked again I think your taste in these matters was exquisite. I really like the distinction you made in the comments between conversion and construction, I'm going to be more deliberate about that in the future.

The override for `issusbset` will become complicated once the backports happen for 0.7, but that's just how it goes.

src/IntervalSets.jl
```diff
@@ -36,4 +74,133 @@ _checked_conversion(::Type{T}, a::T, b::T) where {T} = a, b
...
+# convert numbers to intervals
+convert(::Type{AbstractInterval}, x::Number) = x..x
+convert(::Type{AbstractInterval{T}}, x::Number) where T =
+    convert(AbstractInterval{T}, convert(AbstractInterval, x))
+convert(::Type{TypedEndpointsInterval{:closed,:closed}}, x::Number) = x..x
+convert(::Type{TypedEndpointsInterval{:closed,:closed,T}}, x::Number) where T =
+    convert(AbstractInterval{T}, convert(AbstractInterval, x))
+convert(::Type{ClosedInterval}, x::Number) = x..x
+convert(::Type{ClosedInterval{T}}, x::Number) where T =
+    convert(AbstractInterval{T}, convert(AbstractInterval, x))
```
hyrodium, line 203: Do we need these conversion methods? x-ref:#97

PR #28, opened 2018-07-16
Add `isless` and make `isequal` consistent with `hash`

andyferris:
The current `isequal` won't work in a `Dict` because it violates the principle that `isequal(a, b)` implies `hash(a, b)`. This is corrected to not special case empty intervals.

I also provide an `isless` function so that intervals can be `sort`ed and so-on.

### DIFF
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -3,7 +3,7 @@ __precompile__()
 module IntervalSets
 
 using Base: @pure
-import Base: eltype, convert, show, in, length, isempty, isequal, issubset, ==, hash,
+import Base: eltype, convert, show, in, length, isempty, isequal, isless, issubset, ==, hash,
              union, intersect, minimum, maximum, extrema, range, ⊇
 
 using Compat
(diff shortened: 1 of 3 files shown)


andyferris:
The alternative of course would be to add a check for empty closed intervals in `hash` (and update `isless` accordingly). However, I think performance of `Dict` and `sort` might be slightly better with the approach here.

timholy:
This came while I was traveling without internet, sorry it didn't get attention until now.

I understand your point about `Dict` performance, but I kind of think that the docstring implies that `isequal` is more "permissive," especially about missing values, than `==`. Since an empty interval is kind of like a missing value, I worry this is counter to expectations.

andyferris:
OK. This seems consistent with how we treat empty ranges, arrays, etc.

rofinn:
This is a pretty old PR, but do we still want to add the check to `hash`? I was also surprised by this as it was causing collisions.

```
julia> OpenInterval{Float64}(1, 3) == ClosedInterval{Float64}(1, 3)
false

julia> hash(OpenInterval{Float64}(1, 3)) == hash(ClosedInterval{Float64}(1, 3))
true
```

PR #35, opened 2018-08-28
add Base.IteratorSize method

kdw503:
Add a Base.IteratorSize method to resolve an error in calling issubset() method in julia-0.7.


timholy:
Worth pointing out that this will work in conjunction with https://github.com/JuliaLang/julia/pull/28871.

It is worth adding a test. Until new versions of Julia are released, the test will fail (unfortunately). You could mark it `@test_broken` right now, and then you or someone can change it to `@test` once Julia gets fixed. Near your test, insert a comment like `# see issue #35` so that anyone reading the source code knows to look here for an explanation.

timholy:
Thanks!

PR #48, opened 2019-02-25
Support for missing values

cstjean:
It's straight-forward, and mirrors the semantics of `missing in []` and `missing in [1,2]`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
   - 0.7
   - 1.0
   - nightly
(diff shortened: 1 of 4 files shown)


cstjean:
The build fails on Julia 0.6 because `missing` was not defined. I could either import (and REQUIRE) `Missings.jl`, or require Julia 0.7

ararslan:
I'd drop 0.6 like a hot potato at this point.

cstjean:
Done.

dlfivefifty:
The drop in coverage is unrelated, so I'll merge this.

PR #50, opened 2019-04-24
Expanded the types duration accepts

yakir12:
This will allow duration to work on any typed interval:
```julia
julia> duration(Second(1)..Second(3))
3 seconds
```


yakir12:
Mmm, I now realize that duration should only make sense for integer intervals, and thus this implementation of duration is problematic. We could just add more methods for Dates.Periods etc, but I think that this could get solved if the subtypes of Dates were subtypes of integers. This is hopefully being discussed [here](https://discourse.julialang.org/t/should-dates-period-be-a-subtype-of-integer/23465?u=yakir12).

PR #51, opened 2019-04-24
Add a method to duration

yakir12:
Allows for things like:
```julia
julia> duration(Second(1)..Second(2))
2 seconds
```

### DIFF
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -200,6 +200,7 @@ valued interval. For example, `duration(0..1)` is 2, while `width(0..1)` is 1.
 """
 duration(A::TypedEndpointsInterval{:closed,:closed,T}) where {T<:Integer} = max(0, Int(A.right - A.left) + 1)
 duration(A::TypedEndpointsInterval{:closed,:closed,Date}) = max(0, Dates.days(A.right - A.left) + 1)
+duration(A::TypedEndpointsInterval{:closed,:closed,T}) where {T<:Dates.Period} = max(zero(T), A.right - A.left + oneunit(T))
 
 include("interval.jl")
 


dlfivefifty:
I assume you mean:
```julia
julia> duration(Second(1)..Second(2))
2 seconds
```

yakir12:
You *assume* correctly ;)

timholy:
And what would
```julia
julia> t1 = Time(Second(1), Millisecond(0))
00:00:01

julia> t2 = Time(Second(2), Millisecond(0))
00:00:02

julia> duration(t1..t2)
??
```
return?

yakir12:
Well, this specific PR only adds methods for `Period`s, not `Time`. But if there is interest then this will work:
```julia
duration(A::IntervalSets.TypedEndpointsInterval{:closed,:closed,Time}) = max(Nanosecond(0), A.right - A.left + Nanosecond(1))

julia> duration(t1..t2)
1000000001 nanoseconds
```

cstjean:
`Second(60)..Second(120) == Minute(1)..Minute(2)` is true at the moment, but they would have different durations.

dlfivefifty:
I've always felt `duration` was an ill-defined concept. I'd vote for removing it from IntervalSets.jl.

dlfivefifty:
More precisely: now that we have open sets, interpreting `Minute(1)..Minute(2)` as containing all of minute 2 is no longer necessary, as we can represent the same concept as: `Interval{:closed,:open}(Minute(1),Minute(3))`. 

So my proposal would be:
1. Remove `duration`.
2. Introduce new notation for `Interval{:closed,:open}(Minute(1),Minute(3))`. Unfortunately a Swift-like `Minute(1) ..< Minute(3)` is not possible...
3. Use the new notation wherever `duration` was used, with `width` performing the same role.

yakir12:
I agree that duration is problematic. But just to answer @cstjean's excellent point, this:
```julia
duration(A::IntervalSets.TypedEndpointsInterval{:closed,:closed,T}) where {T<:Dates.Period} = max(0, Dates.value(Nanosecond(A.right - A.left)) + 1)
``` 
would solve that. Basically, just as `Day` is the unit of `Date`, `Nanosecond` is the unit of `Time` and its `Period`s. So just like in the implementation for `Date` arguments, I converted things to `Nanosecond` and then calculated the duration. So there is a solution, but I totally agree that it's kind of arbitrary that the duration of say `Minute(1)..Minute(2)` is `60000000001` (what? minutes? no. milliseconds? no. It's nanoseconds and you should know that somehow). The ambiguity about dates and days is less jarring since everyone will assume that the units of the duration would be `Day`. But still.

That "adding one" is just hard to expect since the units of that "one" are not completely clear.

hyrodium:
I'll close this PR because #102 was merged.

PR #54, opened 2019-09-12
Add copy

dlfivefifty:
This PR adds a copy method. The easiest way to do this was to call `deepcopy`, @ararslan @timholy is there any hidden issue in doing this?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -210,6 +210,10 @@ struct IncompleteInterval <: AbstractInterval{Int} end
 
         @test promote_type(Interval{:closed,:open,Float64}, Interval{:closed,:open,Int}) ===
                         Interval{:closed,:open,Float64}
+
+        J = BigFloat(1)..2                        
+        @test copy(J) == J
+        @test copy(J) !== J
     end
 
 


timholy:
BigFloat is supposed to be semantically immutable but their implementation doesn't let us do that now. `copy(1:2)` has to work because it's part of the AbstractArray interface. Given that `copy(3)` works then I think it's good to define it for intervals too.

`deepcopy` can be a huge performance trap, so I'm a bit opposed to implementing `copy` via `deepcopy`. What problem did using `deepcopy` solve?

dlfivefifty:
I don’t think it actually solves a problem. I’m inclined to change it to
```Julia
copy(d::Domain) = d
```

tkf:
Isn't defining `copy` for concrete subtypes like `Interval` much safer option?  If it's too cumbersome, how about

```julia
copy(d::Domain) = isbits(d) ? d : deepcopy(d)
```

dlfivefifty:
You are certainly right, at least in theory. In practice, I can’t think of a case where it matters. 

But for the sake of not intentionally writing “bad“ code I’ll make that change and only override for ` TypedEndpointsInterval `.

Issue #59, opened 2020-02-10
Rename `isopen`/`isclosed` to `isopenset`/`isclosedset`


daanhb:
The package defines both `isclosed` and `isopen` for intervals, but only `isclosed` is exported.

Using this package downstream and extending `isopen` is complicated by the fact that Base defines `isopen`, albeit with an entirely different meaning in I/O, but IntervalSets does not import it.

The solution may be to both import (from Base) and export `isopen` in IntervalSets? Or is this intentional?

dlfivefifty:
We could import `Base.isopen`, the only reason we didn't is because it referred to a file being open. But on second thought I think its OK for functions to mean different things in different contexts.

daanhb:
It does look odd, but the different meanings would be clearly distinguished by the type of the argument.

The concrete issue for me was that writing `import IntervalSets: isopen` in `DomainSets` yields a warning because of the existence in Base. Yet, not exporting `isopen` ourselves makes the functionality less useful. I'll make a PR.

dlfivefifty:
Probably the best thing to do is make a PR and see if anyone comments

timholy:
Of course you can always say `IntervalSets.isopen(s)`.

I am on the fence about whether `isopen` is a name or a concept; if it's the former, there is no harm extending the Base version. If it's the latter, then very different uses should probably be different `isopen`s (scoped by module). Does anyone know of any guidance about this?

dlfivefifty:
I think it’s a “name”, in the same way * means both multiplication and string concatenation.

daanhb:
Now I see that I misunderstood the situation, sorry. DomainSets was actually working as intended: it imports the `isopen` from IntervalSets and does not export it. Indeed, it is perfectly usable by qualifying with the package name `IntervalSets.isopen` or `DomainSets.isopen`.

Given that apparently I've not used this functionality in a while, it is not exactly a pressing issue to me. I tend to agree that in this case it is a name, and importing `isopen` from Base would be fairly safe. However, this may not be a general solution. It would be interesting to look for other examples.

daanhb:
There is some discussion [here](https://discourse.julialang.org/t/extending-base-methods-with-same-name-completely-different-meaning/11082/3) in favour of keeping functions with different meanings separate, either by renaming or by scoping.

daanhb:
[...]

PR #60, opened 2020-02-11
Rename isopen/isclosed to isopenset/isclosedset

daanhb:
Following the discussion in #59, this pull request adds documentation for the `isopen` function defined by IntervalSets. This function is separate from the existing `isopen` function in Base and it did not yet have documentation.

The pull request adds documentation to the function itself, as well as a brief note to README.md in the section on open intervals. It is an alternative to importing and extending `isopen` from Base.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "IntervalSets"
 uuid = "8197267c-284f-5f27-9208-e0e47529a953"
-version = "0.4.0"
+version = "0.4.1"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "IntervalSets"
 uuid = "8197267c-284f-5f27-9208-e0e47529a953"
-version = "0.4.1"
+version = "0.5.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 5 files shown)


daanhb:
After discussion in #59, the new proposal is to change `isopen` and `isclosed` to `isopenset` and `isclosedset`. I've updated the README with examples, and changed the part where it said that `IntervalSets` defines only the concrete type `ClosedInterval` to reflect the current situation where it defines one concrete type `Interval`. Version number is bumped to 0.4.1.

Should there be a deprecation warning for isopen or isclosed, or for both? Something like:
```julia
@deprecate isopen(d) isopenset(d) false
@deprecate isclosed(d) isclosedset(d)
```
Anything else to think about?

dlfivefifty:
Definitely deprecate. Adding deprecations warrants a bump to v0.5

daanhb:
Okay, it seems deprecation can be done without actually exporting the symbol. Neat:
```julia
julia> using IntervalSets

julia> isclosed(0..1)
┌ Warning: `isclosed(d)` is deprecated, use `isclosedset(d)` instead.
│   caller = top-level scope at REPL[4]:1
└ @ Core REPL[4]:1
true

julia> IntervalSets.isopen(0..1)
┌ Warning: `isopen(d)` is deprecated, use `isopenset(d)` instead.
│   caller = top-level scope at REPL[5]:1
└ @ Core REPL[5]:1
false

julia> isopen(0..1)
ERROR: MethodError: no method matching isopen(::Interval{:closed,:closed,Int64})
```

timholy:
Thanks @daanhb, sorry this took so long!

Issue #61, opened 2020-03-11
`OpenInterval(0,1) ⊆ OpenInterval(0,1)` returns false


hyrodium:
In mathematics, an open interval (a,b) is a subset of itself.
But the code below returns `false`.
```julia
julia> using IntervalSets

julia> OpenInterval(0,1) ⊆ OpenInterval(0,1)
false
```
I think it should return `true`, and this is because of the code below:
```julia
issubset(A::AbstractInterval, B::AbstractInterval) = ((leftendpoint(A) in B) && (rightendpoint(A) in B)) || isempty(A)
```
(from 96th line of https://github.com/JuliaMath/IntervalSets.jl/blob/master/src/IntervalSets.jl)

dlfivefifty:
You are correct. Can you submit a PR?

hyrodium:
Thank you for your comment. I'll try!

Issue #29, opened 2018-07-16
`in(interval, interval)` doesn't make sense (to me)


andyferris:
Julia generally clearly seperates the concepts of containers and their elements.

It seems valid that a interval may be a subset of another interval via `issubset` but generally (scalar) elements will be `in` an interval. I feel it might be preferable to remove the `in(a::ClosedInterval, b::ClosedInterval)` method.

timholy:
Fixed by #62.

PR #62, opened 2020-03-20
Fix issubset function for open interval #61

hyrodium:
This change provides:
```
Interval{:closed,:closed}(1,2) ⊆ Interval{:closed,:closed}(1,2) # true  (previoussly, it returns true )
Interval{:open,  :open  }(1,2) ⊆ Interval{:open  ,:open  }(1,2) # true  (previoussly, it returns false)
Interval{:closed,:open  }(1,2) ⊆ Interval{:open  ,:open  }(1,2) # false (previoussly, it returns false)
Interval{:closed,:open  }(1,2) ⊆ Interval{:closed,:open  }(1,2) # true  (previoussly, it returns false)
```

### DIFF
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -131,22 +131,15 @@ in(::Missing, I::TypedEndpointsInterval{:open,:open}) = !isempty(I) && missing
 in(::Missing, I::TypedEndpointsInterval{:closed,:open}) = !isempty(I) && missing
 in(::Missing, I::TypedEndpointsInterval{:open,:closed}) = !isempty(I) && missing
 
-in(a::AbstractInterval,                         b::TypedEndpointsInterval{:closed,:closed}) =
-    (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
-in(a::TypedEndpointsInterval{:open,:open},      b::TypedEndpointsInterval{:open,:open}) =
-    (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
-in(a::TypedEndpointsInterval{:closed,:open},    b::TypedEndpointsInterval{:open,:open}) =
-    (leftendpoint(a) > leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
-in(a::TypedEndpointsInterval{:open,:closed},    b::TypedEndpointsInterval{:open,:open}) =
-    (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) < rightendpoint(b))
-in(a::TypedEndpointsInterval{:closed,:closed},  b::TypedEndpointsInterval{:open,:open}) =
-    (leftendpoint(a) > leftendpoint(b)) & (rightendpoint(a) < rightendpoint(b))
-in(a::TypedEndpointsInterval{:closed},          b::TypedEndpointsInterval{:open,:closed}) =
-    (leftendpoint(a) > leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
-in(a::TypedEndpointsInterval{:open},            b::TypedEndpointsInterval{:open,:closed}) =
-    (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
-in(a::TypedEndpointsInterval{L,:closed}, b::TypedEndpointsInterval{:closed,:open}) where L = (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) < rightendpoint(b))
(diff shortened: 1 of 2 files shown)


hyrodium:
Sorry, I found mistakes in my pull request.
I'll correct my code.

hyrodium:
I found that the function `in(A:TypedEndpointsInterval,B:TypedEndpointsInterval)` in IntervalSets.jl was what I thought.
For example, the function behave like this:

```julia
I=1..2
OpenInterval(I)  ∈ OpenInterval(I) # true
```

But, I really agree with #29, and the `in` function has a problem when the subset is emptyset:

```julia
I=2..1
J=3..4
OpenInterval(I) ∈ OpenInterval(J) # false
```

So, I changed the `in` function to deprecated, and made the `issubset` function treats like original `in` function (and removed the bug).

The original `issubset` function was defined for `AbstractInterval`, but its implementation consists of `isempty` which treats only `TypedEndpointsInterval`.
So, I also changed the function `issubset` takes only `TypedEndpointsInterval`.

src/IntervalSets.jl
```diff
@@ -137,22 +131,15 @@ in(::Missing, I::TypedEndpointsInterval{:open,:open}) = !isempty(I) && missing
...
-    (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) < rightendpoint(b))
-in(a::TypedEndpointsInterval{:closed,:closed},  b::TypedEndpointsInterval{:open,:open}) =
-    (leftendpoint(a) > leftendpoint(b)) & (rightendpoint(a) < rightendpoint(b))
-in(a::TypedEndpointsInterval{:closed},          b::TypedEndpointsInterval{:open,:closed}) =
-    (leftendpoint(a) > leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
-in(a::TypedEndpointsInterval{:open},            b::TypedEndpointsInterval{:open,:closed}) =
-    (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
-in(a::TypedEndpointsInterval{L,:closed}, b::TypedEndpointsInterval{:closed,:open}) where L = (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) < rightendpoint(b))
-in(a::TypedEndpointsInterval{L,:open}, b::TypedEndpointsInterval{:closed,:open}) where L = (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
+@deprecate in(a::AbstractInterval,                         b::TypedEndpointsInterval{:closed,:closed}) (leftendpoint(a) ≥ leftendpoint(b)) & (rightendpoint(a) ≤ rightendpoint(b))
```
timholy, line 134: In an ideal world, the deprecation warning would direct future users to `issubset`. Could we do these with a manual call to `Base.depwarn`? Maybe something like this:
```julia
function in(a::AbstractInterval, b::AbstractInterval)
    Base.depwarn("`in(a, b)` (equivalently, `a ∈ b`) is deprecated in favor of `issubset(a, b)` (equivalently, `a ⊆ b`). Note that the behavior for empty intervals is also changing.", :in)
    return in_deprecation(a, b)
end
```
and then define `in_deprecation` as above.

src/IntervalSets.jl
```diff
@@ -163,6 +150,32 @@ isequal(A::TypedEndpointsInterval, B::TypedEndpointsInterval) = isempty(A) & ise
 ==(A::TypedEndpointsInterval{L,R}, B::TypedEndpointsInterval{L,R}) where {L,R} = (leftendpoint(A) == leftendpoint(B) && rightendpoint(A) == rightendpoint(B)) || (isempty(A) && isempty(B))
 ==(A::TypedEndpointsInterval, B::TypedEndpointsInterval) = isempty(A) && isempty(B)
 
+function issubset(A::TypedEndpointsInterval, B::TypedEndpointsInterval)
+    Al, Ar = endpoints(A)
+    Bl, Br = endpoints(B)
+    return isempty(A) || ( Bl ≤ Al && Ar ≤ Br )
```
timholy, line 156:
```suggestion
    return isempty(A) | ( Bl ≤ Al && Ar ≤ Br )
```
(the branch in the original is probably actively harmful for performance)

Similar changes below.

hyrodium: Thnaks for the suggestion!
I guess this code would be better, is it right?
```
return isempty(A) | ( Bl ≤ Al & Ar ≤ Br )
```

src/IntervalSets.jl
hyrodium: Thanks for the advice!  I didn't know the function `Base.depwarn`.  I updated the code as you suggest, and pushed the code.

timholy:
Thanks so much, @hyrodium! Excellent contribution.

PR #75, opened 2021-02-04
one->oneunit for Julia v1.6 deprecation

dlfivefifty:


dlfivefifty:
@timholy any idea how to deal with the Julia 1.0 error? (My vote would be to just drop support for < v1.5 instead of trying to debug this...)

timholy:
That sounds like a good fix.

Can't help but notice that this method is weird though. Input a `CartesianIndex`, get a tuple? That's strange.

timholy:
Also, we don't support this with `+`, why should we with `±`?

```julia
julia> i = CartesianIndex(3, 8)
CartesianIndex(3, 8)

julia> i + 1
ERROR: MethodError: no method matching +(::CartesianIndex{2}, ::Int64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:540
  +(::T, ::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8} at int.jl:87
  +(::Rational, ::Integer) at rational.jl:288
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[2]:1

julia> i + oneunit(i)
CartesianIndex(4, 9)
```
That's deliberate. `CartesianIndex` can't be thought of as a container (i.e., iterable) if it's to be able to be equivalent to passing multiple scalar indices. Otherwise broadcasting would try to broadcast over the elements of the CartesianIndex rather than using it as, well, a CartesianIndex. Very important distinction to retain.

dlfivefifty:
@daanhb I suspect this code was copied over from DomainSets.jl. Do you know why this is needed?

hyrodium:
The method `±(x::CartesianIndex, y)` will be removed in the next breaking release, so I'll close this PR. (#101)

Issue #43, opened 2018-10-04
Add range(::ClosedInterval, length=n)


dlfivefifty:
This would be convenient to have:
```julia
range(-1..1, length=10)
```

timholy:
Meaning, it constructs the equivalent of `range(-1, stop=1, length=10)`, right?

myrddin89:
Following up. Is this going to implemented? I would propose to extend the interface like so:

```julia
range(i::ClosedInterval, len) = range(i, length=len)
range(i::ClosedInterval; length) = range(endpoints(i)...; length)
```

I am not sure how to deal with the `step` case. Maybe consider the interval (right)open?
Definitely it could be useful to have a uniform interface for ranges supporting also intervals.

dlfivefifty:
If you make a PR I can merge it

myrddin89:
Sure, will do!

PR #76, opened 2021-03-13
Range interface for Intervals

myrddin89:
This adds the possibility to create a `range` from a `ClosedInterval` or half open interval `Interval{:closed,:open}`.

For a closed interval `[a,b]` this constructs a `range` of length `length` or step `step` equivalent to `range(a, b; step, length)`.

For a half open interval `[a,b)` we do not want `b` to be included, but the range should have the specified `length`, so the step should be `(b-a)/length`.

So for closed intervals the meaning is to create a range starting from `a` and at most ending at `b`;
for half-open intervals this enforces the range to have the given length with step `(b-a)/length`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "IntervalSets"
 uuid = "8197267c-284f-5f27-9208-e0e47529a953"
-version = "0.5.2"
+version = "0.5.3"
 
 [deps]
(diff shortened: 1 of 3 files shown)


myrddin89:
Sorry, forgot to mention #43 for related comments.

dlfivefifty:
Looks good. If you bump the patch version I can merge and tag

PR #78, opened 2021-05-17
support units in findall

aplavin:
units, date, whatever else :)


dlfivefifty:
Can you fix the tests? We can certainly drop Julia v0.7 if it helps

aplavin:
Looks like there was just no division for date periods in 1.0: error says `MethodError: no method matching cld(::Day, ::Day)`. This method was added in a later julia version. So, I'll just remove the test with dates, keeping units and datetimes.

Issue #79, opened 2021-05-18
Examples in Docstring are not updated to Julia1.6


kanak8278:

hyrodium:
This issue was solved with #80.

PR #80, opened 2021-05-18
Updated docstring src/findall.jl

kanak8278:
PR updates the Docstring with the Julia1.6 version. It resolves issue #79.


kanak8278:
@timholy If you would verify my PR.

timholy:
Thanks!

PR #57, opened 2019-11-27
use .. from EllipsisNotation.jl

mcabbott:
Closes #18

PR #83, opened 2021-10-21
Remove EllipsisNotation.jl

mcabbott:
This reverts #57, making the package depend only on the standard library.


mcabbott:
Bump?

mcabbott:
Bump?

timholy:
Seems reasonable. This will definitely be breaking though.

PR #85, opened 2021-11-30
support ± for CartesianIndex

johnnychen94:
In Base we have `CartesianIndex(a, b):CartesianIndex(c,d)` construct
the region `CartesianIndices((a:c, b:d))`. Similiarly, we would
want to have `CartesianIndex(a,b)..CartesianIndex(c,d)` mean the
same thing.

This commit fixes the ± constructor for CartesianIndex

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -97,6 +97,7 @@ Construct a ClosedInterval `iv` spanning the region from
 """
 ±(x, y) = ClosedInterval(x - y, x + y)
 ±(x::CartesianIndex, y) = (xy = y * one(x); map(ClosedInterval, (x - xy).I, (x + xy).I))
+±(x::CartesianIndex, y::CartesianIndex) = ClosedInterval(x-y, x+y)
 
 show(io::IO, I::ClosedInterval) = print(io, leftendpoint(I), "..", rightendpoint(I))
 show(io::IO, I::OpenInterval) = print(io, leftendpoint(I), "..", rightendpoint(I), " (open)")
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -53,6 +53,10 @@ struct IncompleteInterval <: AbstractInterval{Int} end
         O = @inferred(CartesianIndex(1, 2, 3, 4) ± 2)
         @test O == (-1..3, 0..4, 1..5, 2..6)
 
+        x, y = CartesianIndex(1, 2, 3, 4), CartesianIndex(1, 2, 3, 4)
+        O = @inferred x±y
+        @test O == ClosedInterval(x-y, x+y)
+
         @test eltype(I) == Int
         @test eltype(M) == Float64
 


mcabbott:
I see there's some existing code to make this work, although I'm not so sure what it's for:
```
julia> CartesianIndex(1,2) ± 3
(-2..4, -1..5)

julia> typeof(ans)
Tuple{ClosedInterval{Int64}, ClosedInterval{Int64}}
```
Note that it doesn't make a `CartesianIndices`, although that could always encode the same information. Instead it is `(1,2) .± 3`. This PR seems to make a third thing, an interval of `CartesianIndex`. What's the logic of these?

> we would want to have CartesianIndex(a,b)..CartesianIndex(c,d) mean the same thing.

But this is not done in this PR, if I understood right. Again I wonder why this case should be like `:`, and not a tuple of intervals, nor an interval of indices. 

Also, I wonder what the motivation is, should some code be generic over CartesianIndices and some other objects, in a way that this makes nicer?

johnnychen94:
> This PR seems to make a third thing, an interval of CartesianIndex. What's the logic of these?

Although it might not be well-defined on what an interval of `CartesianIndex` is. This PR tries to make it consistent if considering `x ± d` is equivalent to `(x-d)..(x+d)` when it comes to `CartesianIndex`:

```julia
julia> x = d = CartesianIndex(1,2)
CartesianIndex(1, 2)

julia> (x-d)..(x+d)
CartesianIndex(0, 0)..CartesianIndex(2, 4)

julia> x ± d # errors on master
CartesianIndex(0, 0)..CartesianIndex(2, 4)
```

As an example of this, [IntegralArrays](https://github.com/JuliaImages/IntegralArrays.jl) differs the behavior by dispatching on the index types: if `CartesianIndices` then it's common-sense indexing; and if `ClosedInterval` of `CartesianIndex` then it's the special "integral" sense. This becomes quite an convenient notation when dealing with multi-dimensional data.

```julia
julia> using IntegralArrays, IntervalSets

julia> A = [1 2 3 4 5;
            10 20 30 40 50;
            100 200 300 400 500]
3×5 Matrix{Int64}:
   1    2    3    4    5
  10   20   30   40   50
 100  200  300  400  500

julia> Ai = IntegralArray(A)
3×5 IntegralArray{Int64, 2, Matrix{Int64}}:
   1    3    6    10    15
  11   33   66   110   165
 111  333  666  1110  1665

julia> sum(A[1:2, 1:2]) == Ai[2, 2]
true

julia> sum(A[1:3, 2:4]) == Ai[1..3, 2..4] 
true
```

Hope this makes a bit sense.

johnnychen94:
I'm okay to let it wait for another person to approve/merge.

timholy:
I'm willing to be that person, as long as you still want this.

johnnychen94:
Oh yes that would be great. I still think we need this semantic in IntegralArrays

hyrodium (approved):

PR #92, opened 2022-03-30
Drop support for  Julia 0.7

hyrodium:
I think we can drop support for Julia 0.7.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         version:
           - '1.0'
+          - '1.6'
           - '1'
 #          - 'nightly'
         os:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,6 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.0'
           - '1.6'
           - '1'
 #          - 'nightly'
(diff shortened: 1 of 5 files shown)


timholy:
I think this will require a minor version bump, although it's possible that Julia 0.7 is exempted.

hyrodium:
I think dropping support is not a breaking change and it's okay with updating the patch version.

hyrodium:
If we update the minor version bump (with major version 0), I'd like to drop the deprecated methods such as `in(a::AbstractInterval, b::TypedEndpointsInterval{:closed,:closed})`.

hyrodium:
Okay, I'll update the minor version and remove the deprecated methods in this PR.

mcabbott:
The logic, I think, is that bumping only the patch version means there are no intermediate versions you can use to later release bugfix supporting older Julia.

I think there is no chance of doing that, and so would vote for `0.5.5`. Making a breaking change just for this seems like imposing unnecessary hassle on everyone who depends on it.

hyrodium:
I think #83 will also be considered breaking.

mcabbott:
May as well do 1.6, I think. Even if nothing needs it yet, it'll save bumping again when someone thinks of a feature that wants it.

timholy:
We might also consider calling this 1.0?

timholy:
Oh yeah. I vote we just rip those methods out, they seem really busted.

timholy:
We can merge this and then separately decide whether to call the next release 0.6 or 1.0.

timholy:
Thanks @hyrodium !

PR #93, opened 2022-04-02
Fix type promotion

### LINKED ISSUE #86: Constructing intervals with incompatible endpoint types
The `Interval` constructor is permissive and allows endpoints with incompatible types:

```julia
using Unitful: m, s
using IntervalSets #: ..

x = 1m .. 5m # sensible
y = 1m .. 2s # nonsense
width(x) # returns 4m of course
width(y) # fails, understandably, due to dimension mismatch
@which ..(1m, 2s) # constructor is too permissive ??
```

Is there an easy way to tighten the constructor signature so that `1m .. 2s` fails immediately?
In some sense what we want here is only to allow endpoints for which `width` is well defined.
A kludge would be to use try/catch with `width` in the constructor function, but I doubt that would be appealing...

hyrodium:
This PR fixes #86.

```julia
julia> using Unitful: m, s

julia> using IntervalSets

julia> x = 1m .. 5m
1 m..5 m

julia> y = 1m .. 2s
ERROR: promotion of types Unitful.Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}} and Unitful.Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}} failed to change any arguments
Stacktrace:
 [1] error(::String, ::String, ::String)
   @ Base ./error.jl:42
 [2] sametype_error(input::Tuple{Unitful.Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Unitful.Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}})
   @ Base ./promotion.jl:374
 [3] not_sametype(x::Tuple{Unitful.Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Unitful.Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}}, y::Tuple{Unitful.Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Unitful.Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}})
   @ Base ./promotion.jl:368
 [4] promote
   @ ./promotion.jl:351 [inlined]
 [5] (ClosedInterval)(left::Unitful.Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, right::Unitful.Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}})
   @ IntervalSets ~/.julia/dev/IntervalSets/src/interval.jl:30
 [6] ..(x::Unitful.Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, y::Unitful.Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}})
   @ IntervalSets ~/.julia/dev/IntervalSets/src/interval.jl:85
 [7] top-level scope
   @ REPL[4]:1
```

Note that we now have a new issue around `missing`. I'll open it soon.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -23,7 +23,7 @@ closedendpoints(I::MyUnitInterval) = (I.isleftclosed,I.isrightclosed)
 struct IncompleteInterval <: AbstractInterval{Int} end
 
 @testset "IntervalSets" begin
-    @test isempty(detect_ambiguities(IntervalSets, Base, Core))
+    @test isempty(detect_ambiguities(IntervalSets))
 
     @test ordered(2, 1) == (1, 2)
     @test ordered(1, 2) == (1, 2)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -32,6 +32,7 @@ struct IncompleteInterval <: AbstractInterval{Int} end
     @testset "Basic Closed Sets" begin
         @test_throws ErrorException :a .. "b"
         @test_throws ErrorException 1 .. missing
+        @test_throws ErrorException 1u"m" .. 2u"s"
         I = 0..3
         @test I === ClosedInterval(0,3) === ClosedInterval{Int}(0,3) ===
                  Interval(0,3)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -30,6 +30,7 @@ struct IncompleteInterval <: AbstractInterval{Int} end
 
     @testset "Basic Closed Sets" begin
         @test_throws ErrorException :a .. "b"
+        @test_throws ErrorException 1 .. missing
         I = 0..3
         @test I === ClosedInterval(0,3) === ClosedInterval{Int}(0,3) ===
                  Interval(0,3)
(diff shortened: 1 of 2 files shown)


JeffFessler:
Now that #94 and #98 seem done, can you rerun the CI for this PR to see if it works OK?
Or does something else need to be done first?

hyrodium:
I think this is ready to merge. I'll wait some approvals.

test/runtests.jl
```diff
@@ -139,6 +141,13 @@ struct IncompleteInterval <: AbstractInterval{Int} end
         @test_throws MethodError duration(1.2..2.4)
     end
 
+    @testset "Unitful interval" begin
+        @test 1.5u"m" in 1u"m" .. 2u"m"
+        @test 1500u"μm" in 1u"mm" .. 1u"m"
+        @test !(500u"μm" in 1u"mm" .. 1u"m")
+        @test 1u"m" .. 2u"m" == 1000u"mm" .. 2000u"mm"
```
JeffFessler, line 148:
```suggestion
        @test 1u"m" .. 2u"m" == 1000u"mm" .. 2000u"mm"
        @test_throws ErrorException 1m .. 2s
```
Would it be worth verifying that dimension mismatch throws an error since that was the original issue in #86?
(Note that I have not actually tested the line of code suggested here - I'm not that good with git...)
Other than possibly this, the rest LGTM.

hyrodium: Thanks for the suggestion! I think the test should be located around `@test_throws ErrorException 1 .. missing`.
I have added the test `@test_throws ErrorException 1u"m" .. 2u"s"`.

Issue #94, opened 2022-04-02
Drop support for `1..missing`?


hyrodium:
On current `master`, `1..missing` is supported and tested in `runtests.jl` https://github.com/JuliaMath/IntervalSets.jl/blob/4a6e746143de0bbdb05854da17c087dab6b3b018/test/runtests.jl#L638-L641

```julia
julia> 1..missing
1..missing
```

However, with PR #93, it now returns an error.

```julia
julia> 1..missing
ERROR: promotion of types Int64 and Missing failed to change any arguments
Stacktrace:
 [1] error(::String, ::String, ::String)
   @ Base ./error.jl:42
 [2] sametype_error(input::Tuple{Int64, Missing})
   @ Base ./promotion.jl:374
 [3] not_sametype(x::Tuple{Int64, Missing}, y::Tuple{Int64, Missing})
   @ Base ./promotion.jl:368
 [4] promote
   @ ./promotion.jl:351 [inlined]
 [5] (ClosedInterval)(left::Int64, right::Missing)
   @ IntervalSets ~/.julia/dev/IntervalSets/src/interval.jl:30
 [6] ..(x::Int64, y::Missing)
   @ IntervalSets ~/.julia/dev/IntervalSets/src/interval.jl:85
 [7] top-level scope
   @ REPL[6]:1
```

I propose that `1..missing` should return error because this is the same behavior as `Base.range`.

```julia
julia> range(1,missing)
ERROR: MethodError: no method matching (::Colon)(::Int64, ::Missing)
Closest candidates are:
  (::Colon)(::T, ::Any, ::T) where T<:Real at ~/julia/julia-1.7.2/share/julia/base/range.jl:41
  (::Colon)(::A, ::Any, ::C) where {A<:Real, C<:Real} at ~/julia/julia-1.7.2/share/julia/base/range.jl:10
  (::Colon)(::T, ::Any, ::T) where T at ~/julia/julia-1.7.2/share/julia/base/range.jl:40
  ...
Stacktrace:
 [1] range_start_stop(start::Int64, stop::Missing)
   @ Base ./range.jl:144
 [2] _range(start::Int64, step::Nothing, stop::Missing, len::Nothing)
   @ Base ./range.jl:127
 [3] range(start::Int64, stop::Missing; length::Nothing, step::Nothing)
   @ Base ./range.jl:111
 [4] range(start::Int64, stop::Missing)
   @ Base ./range.jl:111
 [5] top-level scope
[...]

PR #98, opened 2022-04-02
Add deprecated message for conversions number to interval

### LINKED ISSUE #97: Drop support for `convert(ClosedInterval,3)`?
`convert(ClosedInterval,3)` is supported on the current master branch.

```julia
julia> convert(ClosedInterval,3)
3..3
```

However, I think this should not be supported like `convert(StepRange,3)`.

```julia
julia> convert(StepRange,3)
ERROR: MethodError: Cannot `convert` an object of type 
  Int64 to an object of type 
  StepRange
Closest candidates are:
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at ~/julia/julia-1.7.2/share/julia/stdlib/v1.7/LinearAlgebra/src/factorization.jl:58
  convert(::Type{T}, ::AbstractRange) where T<:AbstractRange at ~/julia/julia-1.7.2/share/julia/base/range.jl:215
  convert(::Type{T}, ::T) where T<:AbstractArray at ~/julia/julia-1.7.2/share/julia/base/abstractarray.jl:16
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

hyrodium:
This PR fixes #97.

I just removed the conversion methods. Should we add deprecation messages instead?
I guess the conversions are probably not well used.

### DIFF
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -242,7 +242,7 @@ include("interval.jl")
 # convert numbers to intervals
 # TODO: These conversions will be removed in the next breaking release (#97)
 function convert(::Type{AbstractInterval}, x::Number)
-    Base.depwarn("`The conversion number to interval will be removed.", :convert)
+    Base.depwarn("`The conversion number to interval will be removed; construct an interval explicitly, e.g., `x..x`.", :convert)
     x..x
 end
 function convert(::Type{AbstractInterval{T}}, x::Number) where T
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -240,15 +240,31 @@ duration(A::TypedEndpointsInterval{:closed,:closed,Date}) = max(0, Dates.days(A.
 include("interval.jl")
 
 # convert numbers to intervals
-convert(::Type{AbstractInterval}, x::Number) = x..x
-convert(::Type{AbstractInterval{T}}, x::Number) where T =
+# TODO: These conversions will be removed in the next breaking release (#97)
+function convert(::Type{AbstractInterval}, x::Number)
+    Base.depwarn("`The conversion number to interval will be removed.", :convert)
+    x..x
+end
+function convert(::Type{AbstractInterval{T}}, x::Number) where T
+    Base.depwarn("`The conversion number to interval will be removed.", :convert)
     convert(AbstractInterval{T}, convert(AbstractInterval, x))
-convert(::Type{TypedEndpointsInterval{:closed,:closed}}, x::Number) = x..x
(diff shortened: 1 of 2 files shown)


timholy:
I would favor a deprecation, that way we don't have to bump the minor version immediately.

dlfivefifty:
I don’t remember writing these conversions. @daanhb was it you? Any comments?

hyrodium:
>I don’t remember writing these conversions. @daanhb was it you? Any comments?

@dlfivefifty
Maybe by you? :smile: 

* https://github.com/JuliaMath/IntervalSets.jl/pull/26/commits/179646f43ca91961f421e8b7dc93cd4354956fce
* https://github.com/JuliaMath/IntervalSets.jl/pull/26#discussion_r841072374

hyrodium:
>I would favor a deprecation, that way we don't have to bump the minor version immediately.

I thought the next release will be v0.7.0 if we resolve #86 with #93. (cf #94)
I propose dropping support for `1..missing` and making `1..missing` throws an error.

We can release `v0.6.1` with deprecation messages for the conversion of a number to an interval, but I think very few people need this deprecation.

(This is not a strong opinion and I'm okay with the deprecation messages.)

hyrodium:
I'll update this PR with the deprecation messages (x-ref https://github.com/JuliaMath/IntervalSets.jl/issues/94#issuecomment-1086903757)

timholy:
I like the idea of the deprecation because the standard Julia package management makes it easy to see how to fix things:
- receive a CompatHelper PR proposing to update the [compat]
- but the PR fails...let's make sure it really works on the current version
- run the tests, see the hint about how to fix it, and profit

daanhb:
> I don’t remember writing these conversions. @daanhb was it you? Any comments?

I don't think I will miss these conversions, at least I don't recall ever using them. In any case there are other ways of converting a number to a domain.

src/IntervalSets.jl
```diff
@@ -240,15 +240,31 @@ duration(A::TypedEndpointsInterval{:closed,:closed,Date}) = max(0, Dates.days(A.
 include("interval.jl")
 
 # convert numbers to intervals
-convert(::Type{AbstractInterval}, x::Number) = x..x
-convert(::Type{AbstractInterval{T}}, x::Number) where T =
+# TODO: These conversions will be removed in the next breaking release (#97)
+function convert(::Type{AbstractInterval}, x::Number)
+    Base.depwarn("`The conversion number to interval will be removed.", :convert)
```
timholy, line 245:
```suggestion
    Base.depwarn("`The conversion number to interval will be removed; construct an interval explicitly, e.g., `x..x`.", :convert)
```

PR #99, opened 2022-04-04
Add deprecated message for `1..missing`

hyrodium:
This PR adds a deprecated message in the `Interval` constructor, without any breaking changes. (#94)

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -27,7 +27,14 @@ mathematical notation, the constructed range is `(left, right)`.
 const OpenInterval{T} = Interval{:open,:open,T}
 
 Interval{L,R,T}(i::AbstractInterval) where {L,R,T} = Interval{L,R,T}(endpoints(i)...)
-Interval{L,R}(left, right) where {L,R} = Interval{L,R,promote_type(typeof(left), typeof(right))}(left,right)
+function Interval{L,R}(left, right) where {L,R}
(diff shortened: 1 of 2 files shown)


src/interval.jl
```diff
@@ -27,7 +27,14 @@ mathematical notation, the constructed range is `(left, right)`.
 const OpenInterval{T} = Interval{:open,:open,T}
 
 Interval{L,R,T}(i::AbstractInterval) where {L,R,T} = Interval{L,R,T}(endpoints(i)...)
-Interval{L,R}(left, right) where {L,R} = Interval{L,R,promote_type(typeof(left), typeof(right))}(left,right)
+function Interval{L,R}(left, right) where {L,R}
+    # TODO: Replace the retrun value with `Interval{L,R}(promote(left,right)...)`. (#93)
+    T = promote_type(typeof(left), typeof(right))
+    if !isconcretetype(T)
+        Base.depwarn("`The input types cannot be promoted to a concrete type.", :Interval)
```
timholy, line 34:
```suggestion
        Base.depwarn("`Promotion to a concrete type failed and will error in the next release; consider constructing this interval as `Interval{L,R,$(typejoin(typeof(left),typeof(right)))}(left, right)`.", :Interval)
```

hyrodium: Thanks for the suggestion!

PR #112, opened 2022-08-15
fix intersect with different types

aplavin:
also add the same test for union, just in case


hyrodium (approved):
LGTM! Thanks for this PR!

PR #114, opened 2022-08-17
implement Base.mod

aplavin:


dlfivefifty:
Note this is inconsistent with the interval arithmetic version of `mod`.... though our intervals are not the same as interval arithmetic.

@timholy do you have opinions whether this is appropriate?

dlfivefifty:
> `mod(x, 10:15)`

This is a very convincing example. I'm in favour of merging this

src/IntervalSets.jl
```diff
@@ -229,6 +229,13 @@ Clamp the scalar `t` such that the result is in the interval `i`.
 clamp(t, i::TypedEndpointsInterval{:closed,:closed}) =
     clamp(t, leftendpoint(i), rightendpoint(i))
 
+"""
+    mod(x, i::ClosedInterval)
+
+Find `y` in the `i` interval such that `x ≡ y (mod n)`, where `n = width(i)`.
+"""
+mod(x, i::TypedEndpointsInterval{:closed,:closed}) = mod(x - leftendpoint(i), width(i)) + leftendpoint(i)
```
hyrodium, line 237: Do we need this method for `TypedEndpointsInterval{:closed,:closed}`? I thought the method should be defined for `TypedEndpointsInterval{:open,:closed}` and `TypedEndpointsInterval{:closed,:open}` because:

* `mod(3, 1..3) != 3` seems confusing.
* ℝ = ⋯ ⊔ (-2,-1] ⊔ (-1,0] ⊔ (0,1] ⊔ (1,2] ⊔ ⋯ holds, but ℝ = ⋯ ⊔ [-2,-1] ⊔ [-1,0] ⊔ [0,1] ⊔ [1,2] ⊔ ⋯ is not correct.

I think it might be okay to have the method for `TypedEndpointsInterval{:closed,:closed}`, but the docstring can be improved with the prior example in `jldoctest`.

hyrodium: I agree with that. Do you intend to implement the methods for half-open intervals in PR? I'm still thinking we would like to have the methods for mathematical correctness. But this can be done in another PR, so we may merge as is or wait for the changes.

src/IntervalSets.jl
```diff
@@ -229,6 +229,13 @@ Clamp the scalar `t` such that the result is in the interval `i`.
 clamp(t, i::TypedEndpointsInterval{:closed,:closed}) =
     clamp(t, leftendpoint(i), rightendpoint(i))
 
+"""
+    mod(x, i::ClosedInterval)
+
+Find `y` in the `i` interval such that `x ≡ y (mod n)`, where `n = width(i)`.
+"""
```
hyrodium, line 236:
````suggestion
Find `y` in the `i` interval such that `x ≡ y (mod n)`, where `n = width(i)`.

# Examples

```jldoctest
julia> I = 2.5..4.5;

julia> mod(3.0, I)
3.0

julia> mod(5.0, I)
3.0

julia> mod(2.5, I)
2.5

julia> mod(4.5, I)  # (a in I) does not imply (a == mod(a, I))
2.5
```
"""
````

hyrodium (approved):
Thanks for the updates!

PR #104, opened 2022-04-10
Add support for `rand(1..2)`

hyrodium:
This PR fixes #95.

```julia
julia> using IntervalSets, Dates

julia> rand(1..2)  # Sampling from uniform distion on 1..2
1.525710842368245

julia> rand(1..2, 3)  # Get three samples
3-element Vector{Float64}:
 1.0470183699383326
 1.406808529012634
 1.2230905556202805

julia> rand(42..42)  # Accepts closed inerval with zero width.
42.0

julia> rand(Interval{:open,:closed}(1,3))  # Only ClosedInterval is supported

ERROR: ArgumentError: Sampler for this object is not defined
Stacktrace:
 [1] Random.Sampler(#unused#::Type{Random.TaskLocalRNG}, sp::Random.SamplerTrivial{Interval{:open, :closed, Int64}, Float64}, #unused#::Val{1})
   @ Random ~/julia/julia-1.7.1/share/julia/stdlib/v1.7/Random/src/Random.jl:146
 [2] Random.Sampler(rng::Random.TaskLocalRNG, x::Random.SamplerTrivial{Interval{:open, :closed, Int64}, Float64}, r::Val{1})
   @ Random ~/julia/julia-1.7.1/share/julia/stdlib/v1.7/Random/src/Random.jl:140
 [3] rand(rng::Random.TaskLocalRNG, X::Random.SamplerTrivial{Interval{:open, :closed, Int64}, Float64}) (repeats 2 times)
   @ Random ~/julia/julia-1.7.1/share/julia/stdlib/v1.7/Random/src/Random.jl:254
 [4] rand(X::Interval{:open, :closed, Int64})
   @ Random ~/julia/julia-1.7.1/share/julia/stdlib/v1.7/Random/src/Random.jl:259
 [5] top-level scope
   @ REPL[5]:1

julia> rand(42..41)  # Error on empty interval
ERROR: ArgumentError: The interval should be non-empty.
Stacktrace:
 [1] rand(rng::Random.TaskLocalRNG, i::Random.SamplerTrivial{ClosedInterval{Int64}, Float64})
   @ IntervalSets ~/.julia/dev/IntervalSets/src/interval.jl:236
 [2] rand
   @ ~/julia/julia-1.7.1/share/julia/stdlib/v1.7/Random/src/Random.jl:254 [inlined]
 [3] rand(X::ClosedInterval{Int64})
[...]

PR #116, opened 2022-09-02
added rand for TypedEndpointsInterval

zsunberg:
This fixes the issue mentioned at the end of JuliaApproximation/DomainSets.jl#112

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -222,11 +222,11 @@ end
 
 # random sampling from interval
 Random.gentype(::Type{Interval{L,R,T}}) where {L,R,T} = float(T)
-function Random.rand(rng::AbstractRNG, i::Random.SamplerTrivial{ClosedInterval{T}}) where {T<:Real}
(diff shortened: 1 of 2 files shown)


src/interval.jl
```diff
@@ -222,11 +222,11 @@ end
 
 # random sampling from interval
 Random.gentype(::Type{Interval{L,R,T}}) where {L,R,T} = float(T)
-function Random.rand(rng::AbstractRNG, i::Random.SamplerTrivial{ClosedInterval{T}}) where {T<:Real}
+function Random.rand(rng::AbstractRNG, i::Random.SamplerTrivial{<:TypedEndpointsInterval{:closed, R, T}}) where {R, T<:Real}
```
hyrodium, line 225: `clamp` doesn't support `TypedEndpointsInterval{:closed,:open}`, so the type parameter `R` should be replaced with `:closed`. See https://github.com/JuliaMath/IntervalSets.jl/pull/104/files#r846726630 for the reason for the `clamp`.

zsunberg: Done!

hyrodium (approved):
LGTM, thanks!

PR #124, opened 2022-10-16
support ConstructionBase

aplavin:
allows for interop with packages listed at https://github.com/JuliaObjects/ConstructionBase.jl
also, makes changing left/right endpoints convenient

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,11 +3,13 @@ uuid = "8197267c-284f-5f27-9208-e0e47529a953"
 version = "0.7.3"
 
 [deps]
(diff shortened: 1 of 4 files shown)


dlfivefifty (changes requested):
I do not think this should be merged: I do not understand the role of ConstructionBase.jl and we do not want to add unnecessary dependencies.

dlfivefifty:
@timholy any thoughts? 

I'm vey far from convinced (we don't want to depend on every "*Base.jl" package anyone comes up) so if it's up to me we would reject this PR. A glue package IntervalSetsConstructionBase.jl can serve this purpose better for those who need both.

Issue #136, opened 2023-01-20
DomainCore package?


daanhb:
I believe there was a suggestion at some point to create a new package that would define a Domain type. I can't seem to find the comment. Anyway, I think at some point it will be a good way forward, hence this issue. No rush I guess.

Possible definitions may be:
1.
```julia
abstract type Domain end
```

2.
```julia
abstract type AbstractDomain end

abstract type Domain{T} <: AbstractDomain end

Base.eltype(::Type{<:Domain{T}}) where T = T
```

3.
```julia
abstract type Domain{T} end

Base.eltype(::Type{<:Domain{T}}) where T = T
```
These possibilities of course relate to the discussion of whether or not a domain has an `eltype`.

My current personal stance is that domains represent discrete or continuous sets (which sets them apart from existing sets in Julia), that sets are expected to have elements, and that it is useful to convey information about that. However, we could ensure that `T=Any` is a perfectly acceptable option, as it is in Julia for arrays:
```julia
julia> ["1", 2]
2-element Vector{Any}:
  "1"
 2
```
and for dictionaries:
```julia
julia> Dict("2"=>1, 2 => "3")
Dict{Any, Any} with 2 entries:
  2   => "3"
  "2" => 1
```
and for sets:
```julia
julia> Set([1, "2"])
Set{Any} with 2 elements:
  "2"
  1
```
In that case, for simplicity I'd favour the third option above.

hyrodium:
>I believe there was a suggestion at some point to create a new package that would define a Domain type.

x-ref: https://github.com/JuliaMath/IntervalSets.jl/issues/106#issuecomment-1332258685 https://github.com/JuliaMath/IntervalSets.jl/pull/126#issuecomment-1332012309

dlfivefifty:
Probably make it plural a la ChainRulesCore.jl and StaticArrays.jl (either DomainSetsCore.jl or DomainsCore.jl)

dlfivefifty:
But if you are doing that I'd propose going all in the interface as opposed to an abstract type. E.g.
```julia
module DomainCore

abstract type DomainStyle end

"""
    IsDomain()

indicates it satisfies the domain interface, i.e. supports `in`.
"""
struct IsDomain <: DomainStyle end
struct NotDomain <: DomainStyle end

DomainStyle(::Type) = NotDomain()
DomainStyle(::Type{<:AbstractSet}) = IsDomain()
[...]

PR #150, opened 2023-07-29
Deprecate methods for `eltype` and add `boundstype`

### LINKED ISSUE #115: Should eltype(1..2) be Float64? (currently it is Int)
It seems quite strange to me that `eltype(1..2)` is `Int` because the elements in the mathematical set [1,2] are real numbers, and real numbers are usually represented by `Float64`.

Happy to make a PR if people support this.

hyrodium:
Closes #115 ~~and #122~~.

We had a long discussion about `eltype`, and I believe removing them is the best. (https://github.com/JuliaMath/IntervalSets.jl/issues/115#issuecomment-1219278427)

* `eltype` should not be defined for non-iterable object
* We can keep maintaining the implementation as a different function.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "IntervalSets"
 uuid = "8197267c-284f-5f27-9208-e0e47529a953"
-version = "0.7.7"
+version = "0.7.8"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -63,7 +63,15 @@ isopenset(d::AbstractInterval) = isleftopen(d) && isrightopen(d)
 
 boundstype(i::AbstractInterval) = boundstype(typeof(i))
 boundstype(::Type{I}) where {I<:AbstractInterval{T}} where T = T
-@deprecate Base.eltype(I::Type{<:AbstractInterval}) boundstype(I) false
+
+@static if VERSION < v"1.9"
+    function Base.eltype(I::Type{<:AbstractInterval})
+        Base.depwarn("`eltype` for `AbstractInterval` will be replaced with `boundstype` in the next breaking release (v0.8.0).", :eltype)
+        boundstype(I)
+    end
+else
+    @deprecate Base.eltype(I::Type{<:AbstractInterval}) boundstype(I) false
+end
 
 convert(::Type{AbstractInterval}, i::AbstractInterval) = i
 convert(::Type{AbstractInterval{T}}, i::AbstractInterval{T}) where T = i
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -62,8 +62,8 @@ isclosedset(d::AbstractInterval) = isleftclosed(d) && isrightclosed(d)
 isopenset(d::AbstractInterval) = isleftopen(d) && isrightopen(d)
 
 boundstype(i::AbstractInterval) = boundstype(typeof(i))
-boundstype(::Type{AbstractInterval{T}}) where {T} = T
 boundstype(::Type{I}) where {I<:AbstractInterval{T}} where T = T
+@deprecate Base.eltype(I::Type{<:AbstractInterval}) boundstype(I) false
 
 convert(::Type{AbstractInterval}, i::AbstractInterval) = i
 convert(::Type{AbstractInterval{T}}, i::AbstractInterval{T}) where T = i
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -63,7 +63,7 @@ isclosedset(d::AbstractInterval) = isleftclosed(d) && isrightclosed(d)
 isopenset(d::AbstractInterval) = isleftopen(d) && isrightopen(d)
 
 eltype(::Type{AbstractInterval{T}}) where {T} = T
-@pure eltype(::Type{I}) where {I<:AbstractInterval} = eltype(supertype(I))
+eltype(::Type{I}) where {I<:AbstractInterval{T}} where T = T
 
 convert(::Type{AbstractInterval}, i::AbstractInterval) = i
 convert(::Type{AbstractInterval{T}}, i::AbstractInterval{T}) where T = i
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -1,7 +1,6 @@
 module IntervalSets
 
-using Base: @pure
-import Base: eltype, convert, show, in, length, isempty, isequal, isapprox, issubset, ==, hash,
+import Base: convert, show, in, length, isempty, isequal, isapprox, issubset, ==, hash,
              union, intersect, minimum, maximum, extrema, range, clamp, mod, float, ⊇, ⊊, ⊋
 
 using Random
@@ -62,8 +61,9 @@ isclosedset(d::AbstractInterval) = isleftclosed(d) && isrightclosed(d)
 "Is the interval open?"
 isopenset(d::AbstractInterval) = isleftopen(d) && isrightopen(d)
 
-eltype(::Type{AbstractInterval{T}}) where {T} = T
-eltype(::Type{I}) where {I<:AbstractInterval{T}} where T = T
+boundstype(i::AbstractInterval) = boundstype(typeof(i))
+boundstype(::Type{AbstractInterval{T}}) where {T} = T
+boundstype(::Type{I}) where {I<:AbstractInterval{T}} where T = T
 
 convert(::Type{AbstractInterval}, i::AbstractInterval) = i
 convert(::Type{AbstractInterval{T}}, i::AbstractInterval{T}) where T = i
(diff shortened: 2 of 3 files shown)


zsunberg:
Does this mean that the endpoints will always need to be the same type? Might someone want to have a 1.3 to pi interval where they do not want pi converted to Float64?

hyrodium:
>Does this mean that the endpoints will always need to be the same type?

Yes, at least for now.

>Might someone want to have a 1.3 to pi interval where they do not want pi converted to Float64?

Yeah, I thought we should not separate the types of endpoints when I created this PR, but this should be done in another PR if someone wants that feature. (I personally would like to avoid that feature, but this should be discussed in other issues or PRs.)

>Please consider the option to simply deprecate `eltype`, so that not to make a breaking release over a trivial change. It's extremely easy in this case, just use `@deprecate`.

Yes! I'm planning to release the next v0.7.x version with `@deprecate` in `v0.7.x` branch, and release the next breaking release from the `master` branch. I just want to get some reviews to `boundstype` in this PR.

hyrodium:
I will merge this PR in a few days if there are no objections.

dlfivefifty:
> so that not to make a breaking release over a trivial change

Warnings basically break all downstream packages, so this should be treated as a breaking change. Leaving in the deprecation for the next version is fine as it helps debug.

zsunberg:
> Warnings basically break all downstream packages, so this should be treated as a breaking change. Leaving in the deprecation for the next version is fine as it helps debug.

@dlfivefifty I am curious what you mean by this. Deprecation warnings are turned off by default except in tests, so deprecating should be treated as a warning that something *will happen in the future*, not that it is already breaking. https://stackoverflow.com/questions/56135768/which-part-of-semver-should-i-bump-when-deprecating-supported-python-version

hyrodium:
>If this PR does have a big impact on DomainSets.jl, then I hope it does not happen this way. I think everybody agreed that we should move the definition of Domain{T} to a new package such as DomainSetsCore.jl, a very related topic, and I strongly feel that should happen first. I'd be happy to propose something in a few weeks.

I agree with having DomainSetsCore.jl.
I thought we may have `AbstractInterval{S} <: DomainSetsCore.Domain` (`S` is `boundstype`) without the type parameter of `Domain`, but is this less informative? If we need more information, is it better to have `Domain{T, N}` which represents the typical element type `T` and the dimension of the domain `N`? (For example, a domain on a sphere may be `<: Domain{SVector{3,Float64}, 2}`)

I don't have practical usage of `Domain` with type parameters for now, so some examples would be preferred. (I will look into DomainSets.jl)

>having an `eltype` to me is clearly way better and simpler than any alternatives

Is it okay to rename `eltype` to `typical_eltype` just for avoiding confusion such as #115? (The name `typical_eltype` is temporary, just for discussion.)

daanhb:
> I agree with having DomainSetsCore.jl. I thought we may have `AbstractInterval{S} <: DomainSetsCore.Domain` (`S` is `boundstype`) without the type parameter of `Domain`, but is this less informative? If we need more information, is it better to have `Domain{T, N}` which represents the typical element type `T` and the dimension of the domain `N`? (For example, a domain on a sphere may be `<: Domain{SVector{3,Float64}, 2}`)

The combination `{T,N}` is less general and flexible than having `T = SVector{N,S}`, because the former is exclusive to Euclidean geometry and the latter isn't - so I think one parameter `T` will do. As one example that is relevant in practice, `T = Vector{S}` allows domains with an arbitrary ("flexible") dimension not encoded in the type, but `{T,N}` doesn't.

This argument only holds for the abstract supertype. A concrete domain could always do `MyEuclideanDomain{T,N} <: Domain{SVector{T,S}}`. Or `MyDomain <: Domain{Any}`, if `eltype` is not relevant.

> I don't have practical usage of `Domain` with type parameters for now, so some examples would be preferred. (I will look into DomainSets.jl)

The simplest examples arise from taking cartesian products:
```julia
julia> using IntervalSets, DomainSets

julia> using DomainSets: ×

julia> d1 = (1..2) × (3..4.0)
(1.0 .. 2.0) × (3.0 .. 4.0)

julia> eltype(d1)
SVector{2, Float64} (alias for StaticArraysCore.SArray{Tuple{2}, Float64, 1, 2})

julia> d2 = (1.0..2.0) × UnitCircle()
(1.0 .. 2.0) × UnitCircle()

julia> eltype(d2)
SVector{3, Float64} (alias for StaticArraysCore.SArray{Tuple{3}, Float64, 1, 3})

julia> d3 = (1.0..2.0) × UnitBall(5)
(1.0 .. 2.0) × UnitBall(5)

julia> eltype(d3)
Tuple{Float64, Vector{Float64}}
```
There is a lot going on here, but these are the sort of things on my mind when I say that `T` conveys "structure" of the domain.

> Is it okay to rename `eltype` to `typical_eltype` just for avoiding confusion such as #115? (The name `typical_eltype` is temporary, just for discussion.)

I feel like `typical_eltype` is more confusing than `eltype`. My guess is that having a single intended `eltype`, for those who care about element types, is the most common case. (There might be other element types, sure, but at least this one works.)

daanhb:
I took a [stab](https://github.com/JuliaApproximation/DomainSetsCore.jl/blob/main/src/DomainSetsCore.jl) at `DomainSetsCore.jl`.

The first iteration has `AbstractDomain` and `Domain{T} <: AbstractDomain`. IntervalSets could do one of the following:
- let AbstractInterval{T} inherit from Domain{T}: in that case T is the eltype
- let AbstractInterval or AbstractInterval{T} inherit from AbstractDomain: in that case the eltype is `Any` (the default eltype of anything)
- ignore DomainSetsCore.jl and let AbstractInterval be the main supertype of this package

In the long run I'd say the third option may be the preferred one. It is also closest to what @dlfivefifty has been pushing for, namely domains as an interface. Ideally, we could get DomainSets.jl to work with Intervals like that, regardless of inheritance of any supertype. Because that would mean that DomainSets could also work with other packages defining domains, with all packages able to evolve independently.

To me, in this design `AbstractDomain` is functionally equivalent to `Domain{Any}` and I think I'd rather remove it.

I've added an `AsDomain` constructor for experimentation. The idea is that one can do `some_function(a, b, AsDomain(c))` to hint that `c` behaves like and should be interpreted as a domain, much like `Ref(x)` is used in broadcast to hint at something. One might then be able to do `1..2 \cup AsDomain([4,5,6])` and get the union of the interval with the vector, without actually ever converting the vector to a wrapper Domain type.

zsunberg:
> ignore DomainSetsCore.jl and let AbstractInterval be the main supertype of this package

@daanhb , I was about to log on and suggest this! I think DomainSets would be more powerful if it relies as little as possible on things being a subtype of `Domain`.

daanhb:
There was already an issue for DomainSetsCore.jl with some discussion: #136. I'll move there.

daanhb:
Meanwhile, a lot of changes happened over in DomainSets (unmerged, but in [this PR](https://github.com/JuliaApproximation/DomainSets.jl/pull/141)), which are beneficial independently. Three possibilities to move forward are:
- we incorporate the changes in DomainSets itself, release v0.7, and IntervalSets.jl does not change
- we move the definition of Domain{T} from IntervalSets.jl to DomainSetsCore.jl and make two new releases. In this case the functionality of IntervalSets.jl remains the same
- the intervals in IntervalSets no longer inherit from a Domain supertype and just adhere to the domain interface. (This will probably require further changes in DomainSets to ensure that things remain easy to use)

I think that, if there is a common Domain supertype, it should be in a separate package. I think that it is not worth pursuing convincing types from other packages to inherit from Domain: making the interface easier to work with is clearly the way to go. I've found that few other domain-like types implement `eltype` the way DomainSets expects it, so working with `domaineltype` (name still up for discussion, preferably something shorter) also seems to way to go.

Based on this, for now I prefer the second option above: we register DomainSetsCore, move the definition of Domain{T} there, and make no other changes. That allows to experiment with domains as an interface, hopefully fix and improve things, and then move on.

As for IntervalSets, I believe the goal should be that the simplest possible sensible definition of an interval, namely
```julia
struct Interval{T}
    a::T
    b::T
end
```
should "just work". That won't work yet with option 2 above. For anyone wishing to do computations with such intervals as a continuous set, a good question to ask is: how does one (i) discover and (ii) control the type being returned by `rand(d)` for an interval in which the endpoints are integers.

zsunberg:
> how does one (i) discover and (ii) control the type being returned by rand(d) for an interval in which the endpoints are integers.

Regarding discovery, there is `Random.gentype`. However, it is not documented, so maybe it should not be used outside of `Base`?

I think that *if* we adopt the struct proposed above, we should just convert every sampled type to the `T` anyways, i.e. `rand(1..2)` should return an `Int`. It might be a bit of a pain to deal with something like the open interval (1,2), but that does not seem like a show stopper. People will quickly start using `1.0..2.0` instead of `1..2`.

For *controlling* the type returned, one option would be
```julia
struct Interval{T, A, B}
    a::A
    b::B
end
```
This has the added benefit of allowing different endpoint types, which could be useful for, e.g. `1..pi`.

hyrodium:
Sorry for the late response.
Let me summarise the current discussions.

* `Base.eltype`
  * Typically, this function is used for iterators.
  * We all agree with that `eltype{1..2} == Int` is problematic, right?
  * I'm not sure `eltype(1..2) == Float` is the way to go. (https://github.com/JuliaMath/IntervalSets.jl/issues/115#issuecomment-1219229766)
* `Random.gentype`
  * This is not well-documented and may be removed in the future. (https://github.com/JuliaLang/julia/issues/31968#issuecomment-499037485)
  * But I think this function should be stable and documented; `eltype` for iterators, `gentype` for random generators.
  * This package defines `gentype(::Type{<:Interval})` because I would like to keep the behavior of `eltype(::Type{<:Interval})` (https://github.com/JuliaMath/IntervalSets.jl/pull/104).
  * https://github.com/JuliaMath/IntervalSets.jl/blob/6545e5da1760907338c0315ce76efeaf915f03a6/src/interval.jl#L162
* `IntervalSets.boundstype`
  * A useful function to obtain the type of the boundaries of an interval.
* `DomainSetsCore.domaineltype`
  * [documentation](https://github.com/JuliaApproximation/DomainSetsCore.jl#the-domaineltype-function)
  * We don't have agreements on the proper definition of "eltype", so let's define our function!

I think the following definitions would be acceptable.

```julia
Base.eltype(::Type{<:Interval})  # This is not defined
Random.gentype(::Type{Interval{L,R,T}}) where {L,R,T} = float(T)
Base.boundstype(::Type{Interval{L,R,T}}) where {L,R,T} = T
DomainSetsCore.domaineltype(::Type{Interval{L,R,T}} = float(T)  # I don't have usecases of this method currently, but this would be useful in DomainSets.jl, right?
```

PR #155, opened 2023-08-22
move Domain to DomainSetsCore

daanhb:
This PR is an example of how the definition of Domain could move to a DomainSetsCore package.

For now, this does not change anything else apart from the move.

The PR can't be merged until DomainSetsCore has been registered.

A related (and much more invasive) PR for the DomainSets package is https://github.com/JuliaApproximation/DomainSets.jl/pull/140.


daanhb:
I'll update this elsewhere

Issue #100, opened 2022-04-04
`findall` throws error for zero-step range


hyrodium:
```julia
julia> r = range(1,1,10)
StepRangeLen(1.0, 0.0, 10)

julia> findall(in(1..3),r)
ERROR: InexactError: trunc(Int64, NaN)
Stacktrace:
 [1] trunc
   @ ./float.jl:805 [inlined]
 [2] round
   @ ./float.jl:369 [inlined]
 [3] findall(interval_d::Base.Fix2{typeof(in), ClosedInterval{Int64}}, x::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64})
   @ IntervalSets ~/.julia/dev/IntervalSets/src/findall.jl:58
 [4] top-level scope
   @ REPL[6]:1
```

PR #168, opened 2023-11-09
use old searchsorted() implementation until julia 1.11

aplavin:
Looks like the recently merged https://github.com/JuliaMath/IntervalSets.jl/pull/111 has a significant overhead on Julia 1.10(rc), 1.9, and earlier. It's only in 1.11 with https://github.com/JuliaLang/julia/pull/50365 where the overhead disappears.
Here, I revert to the old `searchsorted_interval` implementation for older Julias.

Not just ranges – probably any type that is not an `Array` (?):
```julia
needle = 1..10

haystack = 1:10^5
@btime searchsorted_interval($haystack, $needle)
# IntervalSets 0.7.7 or this PR:
  2.541 ns (0 allocations: 0 bytes)
# IntervalSets 0.7.8:
  83.808 ns (2 allocations: 64 bytes)

using FlexiMaps
haystack = mapview(x -> x, collect(1:10^5))
@btime searchsorted_interval($haystack, $needle)
# IntervalSets 0.7.7 or this PR:
   50.997 ns (0 allocations: 0 bytes)
# IntervalSets 0.7.8:
  123.564 ns (2 allocations: 32 bytes)
```
For a package that's supposed to be used in all kinds of code, including performance-sensitive, these are highly noticeable.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "IntervalSets"
 uuid = "8197267c-284f-5f27-9208-e0e47529a953"
-version = "0.7.8"
+version = "0.7.9"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"


hyrodium:
Reverting the function may break the zero-step range test (https://github.com/JuliaMath/IntervalSets.jl/issues/100), right?

I think adding the following description to the documentation would be reasonable.

---

There was a performance degression in IntervalSets.jl v0.7.8 because of a Base implementation (https://github.com/JuliaLang/julia/pull/50365).

- For users who really care about the performance
  - and would like to use the latest version of this package:
    - Use upcoming Julia v1.11.
  - and are using Julia earlier than v1.11:
    - Install this package with `]add IntervalSets#v0.7.7`
    - Or, add `IntervalSets = "0.7.7"` to `compat` table in your `Project.toml`.
- For users who don't care about the performance:
  - Install the latest version of IntervalSets.jl on your Julia environment as usual.

dlfivefifty:
I agree with @hyrodium: trying to keep performance high on _every_ supported version of Julia is going to add a lot of burden to both the code writer and maintainers. 

Is there evidence that this code is impacting other packages out in the wild? If not I think if its fine if its slow until Julia v1.11.

dlfivefifty:
Ah ok thanks for the explanation! Yes I agree if its impacting your code we should def fix it

src/findall.jl
```diff
@@ -65,15 +110,24 @@ julia> searchsorted_interval(Float64[], 1..3)
...
-        _searchsorted_begin(X, rightendpoint(i), Val(R); rev):_searchsorted_end(X, leftendpoint(i), Val(L); rev)
-    else
-        _searchsorted_begin(X, leftendpoint(i), Val(L); rev):_searchsorted_end(X, rightendpoint(i), Val(R); rev)
+function searchsorted_interval end
+
+if VERSION < v"1.11-"
+    searchsorted_interval(X, i::Interval{:closed, :closed}) = searchsortedfirst(X, leftendpoint(i))     :searchsortedlast(X, rightendpoint(i))
+    searchsorted_interval(X, i::Interval{:closed,   :open}) = searchsortedfirst(X, leftendpoint(i))     :(searchsortedfirst(X, rightendpoint(i)) - 1)
+    searchsorted_interval(X, i::Interval{  :open, :closed}) = (searchsortedlast(X, leftendpoint(i)) + 1):searchsortedlast(X, rightendpoint(i))
+    searchsorted_interval(X, i::Interval{  :open,   :open}) = (searchsortedlast(X, leftendpoint(i)) + 1):(searchsortedfirst(X, rightendpoint(i)) - 1)
```
hyrodium, line 119: If we remove the keyword argument `rev`, that would be breaking. Could you add the keyword argument, or bump the version to `v0.8.0`?

src/findall.jl
```diff
@@ -65,15 +110,24 @@ julia> searchsorted_interval(Float64[], 1..3)
...
+else
+    function searchsorted_interval(X, i::Interval{L, R}; rev=false) where {L, R}
+        if rev === true
+            _searchsorted_begin(X, rightendpoint(i), Val(R); rev):_searchsorted_end(X, leftendpoint(i), Val(L); rev)
```
hyrodium, line 123: This line is not tested on CI. Could you remove this `#` in `CI.yml`?

https://github.com/JuliaMath/IntervalSets.jl/blob/c1f78d6d4ec8f3c516448a3be47ff01e56d0b541/.github/workflows/CI.yml#L18

hyrodium:
I'm not sure that reopening the issue with Unitful.jl is considered as breaking.
https://github.com/JuliaMath/IntervalSets.jl/pull/111#issuecomment-1767598579

src/findall.jl
hyrodium: I understand the situation. If we revert #111, then we lose `rev`, and that must be considered as a breaking change. If you would like to avoid the breaking release, this PR should implement `rev` keyword argument without performance regression.

I'm okay with both choices: breaking release or adding the `rev` argument.

PR #171, opened 2024-01-03
add geomean(interval) + reduce deps

aplavin:
`Statistics.mean(interval)` already works, so it's only natural to add `StatsBase.geomean(interval)` – that's what I do here.

Also reduce deps on stdlibs:
- Dates are just not needed, I simply removed `using Dates`
- Made `Random` a weakdep on 1.9+

So, now IntervalSets is completely dependency-free (on 1.9+), even regarding stdlibs!


src/IntervalSets.jl
```diff
@@ -283,6 +279,7 @@ include("findall.jl")
 
 if !isdefined(Base, :get_extension)
     include("../ext/IntervalSetsStatisticsExt.jl")
+    include("../ext/IntervalSetRandomExt.jl")
```
hyrodium, line 282:
```suggestion
    include("../ext/IntervalSetsRandomExt.jl")
    # We don't support the extension with StatsBase.jl before Julia v1.9.
    # include("../ext/IntervalSetsStatsBaseExt.jl")
```
* `IntervalSetRandomExt` should be `IntervalSetsRandomExt`.
* Adding the above comments will help other developers.

hyrodium (commented):
Thanks for the PR!
Just added minor comments. Other changes look good to me.

hyrodium (approved):
I'll merge this PR in a few days if there are no objections.

PR #177, opened 2024-02-04
Add `@iv_str` macro

### LINKED ISSUE #39: Alternative to specify open/closed endpoints

hyrodium:
This PR fixes #39.

```julia
julia> using Test, IntervalSets

julia> iv"[1,2]"
1 .. 2

julia> iv"[1,2)"
1 .. 2 (closed-open)

julia> iv"(1,2]"
1 .. 2 (open-closed)

julia> iv"(1,2)"
1 .. 2 (open)
```

### DIFF
--- a/src/IntervalSets.jl
+++ b/src/IntervalSets.jl
@@ -4,7 +4,7 @@ using Base: @pure
 import Base: eltype, convert, show, in, length, isempty, isequal, isapprox, issubset, ==, hash,
              union, intersect, minimum, maximum, extrema, range, clamp, mod, float, ⊇, ⊊, ⊋
 
-export AbstractInterval, Interval, OpenInterval, ClosedInterval,
+export AbstractInterval, Interval, OpenInterval, ClosedInterval, @iv_str,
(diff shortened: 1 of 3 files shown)


hyrodium:
I will merge this PR in a few days if there are no objections.

PR #182, opened 2024-02-08
Add plot recipe

hyrodium:
```julia
julia> using IntervalSets, Plots

julia> plot(iv"(1,2)"); plot!(iv"[3,6)"); plot!(iv"[5,7)")
```

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .DS_Store
 Manifest.toml
 /docs/build/
+/test/out_plot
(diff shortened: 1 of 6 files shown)


hyrodium:
Note that the open endpoints are not transparent.

```julia
julia> using IntervalSets, Plots

julia> plot(iv"[1,3]"); plot!(iv"(2,4)")
```

image

The `offset` keyword argument is useful for better visualization.

```julia
julia> using IntervalSets, Plots

julia> plot(iv"[1,3]"); plot!(iv"(2,4)"; offset=-0.1, ylims=(-1,1))
```

image

PR #188, opened 2024-02-14
Add `ENV["GKSwstype"] = "100"` in `make.jl`

hyrodium:
This PR fixes the following warning message.

>GKS: cannot open display - headless operation mode active

image

ref: https://github.com/JuliaPlots/Plots.jl/issues/1076#issuecomment-327509819


hyrodium:
Fixed! :tada:

image

PR #193, opened 2024-09-26
Provide `range` methods for `OpenInterval` and `Interval{:open,:closed}`

Sagnac:
Methods already exist for `ClosedInterval` and `Interval{:closed,:open}`; these new methods complete the set of range extensions for each interval type.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - 'lts'
           - '1'
           - 'nightly'
         os:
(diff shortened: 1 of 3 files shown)


Sagnac:
@hyrodium it looks like the Julia 1.6 tests are failing because that version of Julia doesn't support all `range` methods; since LTS is now at Julia 1.10.5 do we still want to support 1.6?

dlfivefifty:
No just change the tests to test 'lts' and '1'

Sagnac:
Done

dlfivefifty:
For some reason JET.jl is v0.8.x but the newest version is v0.9.x

hyrodium (approved):
Thank you for the PR! The test failure with Invalidations should be unrelated to this PR.

PR #196, opened 2024-11-25
add min/maximum(abs)

aplavin:


src/IntervalSets.jl
```diff
@@ -337,6 +337,12 @@ maximum(d::TypedEndpointsInterval{L,:open}) where L = throw(ArgumentError("$d is
 
 extrema(I::TypedEndpointsInterval) = (infimum(I), supremum(I))
 
+minimum(::typeof(abs), I::TypedEndpointsInterval{:closed,:closed}) =
+    minimum(I) < 0 < maximum(I) ?
+        zero(minimum(I)) :
```
cstjean, line 342:
```suggestion
minimum(::typeof(abs), I::TypedEndpointsInterval{:closed,:closed,T}) where T=
    zero(T) in I ?
        zero(T) :
```
It's a rather obscure point of implementation, but since eg. intervals of DateTime are supported, `zero` might be more correct for some such type.

Apart from that LGTM.

test/base_methods.jl
```diff
@@ -102,3 +102,23 @@ end
...
+    @test maximum(abs, 1u"m"..2u"m") == 2u"m"
+    @test maximum(abs, -3u"m"..2u"m") == 3u"m"
+    @test minimum(abs, 1u"m"..2u"m") == 1u"m"
+    @test minimum(abs, -3u"m"..2u"m") == 0u"m"
+
+    @test extrema(abs, 1..2) == (1, 2)
+    @test extrema(abs, -3..2) == (0, 3)
+    @test_throws Exception extrema(abs, iv"[1, 2)")
+    @test extrema(abs, iv"[-3, 2)") == (0, 3)
+end
```
hyrodium, line 124: This might be a bit tedious, but could you add tests like this?
```suggestion
    @test extrema(abs, iv"[-3, 2)") == (0, 3)

    @test_throws ArgumentError minimum(abs, 2..1)
    @test_throws ArgumentError minimum(abs, iv"(1,2)")
    @test_throws ArgumentError minimum(abs, iv"[-3,-1)")
    @test_throws ArgumentError minimum(abs, iv"(1,3]")

    @test_throws ArgumentError maximum(abs, 2..1)
    @test_throws ArgumentError maximum(abs, iv"(1,2)")
    @test_throws ArgumentError maximum(abs, iv"(-3,-1]")
    @test_throws ArgumentError maximum(abs, iv"[1,3)")
end
```

hyrodium (commented):
I just added a comment to improve the test coverage. Other changes look great to me!

hyrodium (changes requested):
Sorry for the late review. I found a edge case.
```julia
julia> minimum(abs, 1..2)
1

julia> minimum(abs, 1..1)  # This should return 1
ERROR: ArgumentError: cannot determine minimum(abs, 1 .. 1)
Stacktrace:
 [1] minimum(::typeof(abs), I::ClosedInterval{Int64})
   @ IntervalSets ~/.julia/dev/IntervalSets/src/IntervalSets.jl:355
 [2] top-level scope
   @ REPL[11]:1
```

aplavin:
probably makes sense to define these elsewhere, there's more useful functions than `abs`

PR #198, opened 2025-02-27
explicitly qualify `UnitRange` while adding method

nsajko:
Prevent name ambiguity warning during precompilation on nightly Julia.

See:
* https://github.com/JuliaLang/julia/issues/25744
* https://github.com/JuliaLang/julia/issues/57290
* https://github.com/JuliaLang/julia/pull/57311

PR #199, opened 2025-03-01
Fix warning on Julia nightly

asinghvi17:
It seems that Julia nightly desires all method definitions of names that have not explicitly been imported to be qualified (with e.g. `$module.$name`.


asinghvi17:
Closing as duplicate of #198

PR #201, opened 2025-07-30
add Printf support

aplavin:
It's very convenient to print intervals (and other objects) with flexible number formatting. Printf is reasonably extensible, and this PR adds Printf support for intervals. + corresponding tests.
FWIW, there are a few packages with this kind of Printf support already: https://juliahub.com/ui/Search?type=code&q=Printf.fmt.


dlfivefifty (approved):

aplavin:
can we get a release with this? :) @hyrodium

